PR #3, opened 2019-05-10
Drop DRF from dependencies

menecio:
DRF was used for deserializing to JSON, this can be achieved by using standard python.

### DIFF
--- a/rele/client.py
+++ b/rele/client.py
@@ -1,10 +1,10 @@
+import json
 import logging
 from contextlib import suppress
 
 from django.conf import settings
 from google.api_core import exceptions
 from google.cloud import pubsub_v1
-from rest_framework.renderers import JSONRenderer
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


menecio:
Depends on #2

rele/client.py
```diff
@@ -49,7 +49,7 @@ def __init__(self):
             credentials=settings.RELE_GC_CREDENTIALS)
 
     def publish(self, topic, data, **attrs):
-        data = JSONRenderer().render(data)
+        data = json.dumps(data).encode()
```
andrewgy8: Im afraid this wont work in the case that you pass in `serializers.data`. We tried this, and thats why we had to use `data = JSONRenderer().render(data)`. If anything, you will would have to pass in your data already encoded to the publish function, or we make the encoding method configurable.

tobami:
To be implemented here: https://github.com/mercadona/rele/issues/8

menecio:
@tobami why did you close this one without having an agreement on how to proceed? Because DRF compat can be achieved right now without doing anything fancy, just by know that you must use `.to_representation()` method when publishing data

rele/client.py
menecio: Did you try passing `serializer.to_representation()` ?

PR #13, opened 2019-05-15
Enhance the make release command with twine

andrewgy8:
### :tophat: What?

Enhance the make release command with twine

### :thinking: Why?

Makes deployments easier


andrewgy8:
I remove the travis deploy step, and instead enhanced the `make release` command. If we find this to be too much we can think about integrating it back into travis. IMHO, right now is not the time to automate this as it will probably cause more trouble than its worth.

andrewgy8:
In addition, I was looking at some other projects that make use of travis (restframework, requests, mode, etc.) and none of them use the `deploy` feature of travis.

PR #14, opened 2019-05-20
Add docstrings to the sub decorator

andrewgy8:
### :tophat: What?

Add docstrings to the sub decorator to get documentation started. 

### :thinking: Why?

Just getting started with documenting how the rele package works. 

Is this how we want our docstrings styled?

### DIFF
--- a/rele/subscription.py
+++ b/rele/subscription.py
@@ -75,10 +75,10 @@ def sub(topic, suffix=None):
         def purpose_2(data, **kwargs):
              pass
 
-    :param topic: string The topic that is being subscribed too.
+    :param topic: string The topic that is being subscribed to.
     :param suffix: string An options suffix to the subscription name.
-    Useful when you have two subscribers in the same project that are subscribed to
-    the same topic.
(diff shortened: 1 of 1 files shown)


rele/subscription.py
```diff
@@ -50,6 +50,37 @@ def __call__(self, message):
...
+
+        @sub(topic='lets-tell-everyone', suffix='sub1')
+        def purpose_1(data, **kwargs):
+             pass
+
+        @sub(topic='lets-tell-everyone', suffix='sub2')
+        def purpose_2(data, **kwargs):
+             pass
+
+    :param topic: string The topic that is being subscribed too.
```
tobami: "... subscribed to"

rele/subscription.py
```diff
@@ -50,6 +50,37 @@ def __call__(self, message):
...
+    :param suffix: string An options suffix to the subscription name.
+    Useful when you have two subscribers in the same project that are subscribed to
```
tobami: This should be indented to the first char of the param description ("string")

PR #16, opened 2019-05-30
Add metrics in subscription log

LluisGuiVi:
### :tophat: What?

Add extra parameters to subscription log

### :thinking: Why?

This allows monitoring by prometheus

### DIFF
--- a/rele/subscription.py
+++ b/rele/subscription.py
@@ -34,7 +34,7 @@ class Callback:
         db.close_old_connections()
 
         logger.info(f'Start processing message for {self._subscription}',
-                    extra=self._build_task_info())
+                    extra=self._build_metrics())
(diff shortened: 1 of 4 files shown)


rele/subscription.py
```diff
@@ -32,7 +33,8 @@ def __init__(self, subscription, suffix=None):
     def __call__(self, message):
         db.close_old_connections()
 
-        logger.info(f'Start processing message for {self._subscription}')
+        logger.info(f'Start processing message for {self._subscription}',
+                    extra=self._build_task_info())
```
tobami: There is no concept of "task" here. I'd call it `_build_metrics()` or `_build_sub_info()`

LluisGuiVi: Done

PR #18, opened 2019-05-30
Add metrics to failed subscription log

LluisGuiVi:
### :tophat: What?

Add metrics to failed subscription log

### :thinking: Why?

To obtain metrics from subscription error

### DIFF
--- a/rele/subscription.py
+++ b/rele/subscription.py
@@ -1,5 +1,6 @@
 import json
 import logging
+from enum import Enum
 
 from django import db
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


rele/subscription.py
```diff
@@ -26,6 +27,13 @@ def __str__(self):
 
 class Callback:
 
+    class SubscriptionStatus(Enum):
```
antoniobusrod: I would suggest to use plain constants instead of extending Enum for constants.

LluisGuiVi:
We are going to close the branch because the metrics structure are going to change

PR #22, opened 2019-05-31
 Add remaining subscription log

jjponz:
### :tophat: What?

Add log information in all subscription process

### :thinking: Why?

To can scrap the metrics in prometheus

### DIFF
--- a/tests/test_subscription.py
+++ b/tests/test_subscription.py
@@ -103,7 +103,7 @@ class TestCallback:
         callback = Callback(sub_stub)
         callback(message_wrapper)
 
-        success_log = caplog.records[1]
+        success_log = caplog.records[-1]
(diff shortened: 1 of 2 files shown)


tests/test_subscription.py
```diff
@@ -70,12 +71,39 @@ def test_acks_message_when_callback_called(self, caplog, message_wrapper):
...
+        with caplog.at_level(logging.DEBUG):
+            callback = Callback(sub_stub)
+            res = callback(message_wrapper)
+
+        assert res is None
+        message_wrapper_log = caplog.records[1]
+        assert message_wrapper_log.message == ('I am a task doing '
+                                               'stuff with ID 123 (es)')
+
+    def test_log_when_callback_when_execution_is_succesful(
```
tobami: `when` appears twice

PR #24, opened 2019-06-03
Change published log format to can scrape this with prometheus

jjponz:
### :tophat: What?

Change published log format to can scrape this with prometheus

### :thinking: Why?

To can use logs in grafana

### DIFF
--- a/rele/client.py
+++ b/rele/client.py
@@ -49,7 +49,11 @@ class Publisher:
 
     def publish(self, topic, data, blocking=False, **attrs):
         payload = JSONRenderer().render(data)
-        self._save_log(attrs, topic)
+        logger.info(f'Publishing to {topic}',
+                    extra={'pubsub_publisher_attrs': attrs,
+                           'metrics': self._build_metrics(topic),
+                           }
+        )
         topic_path = self._client.topic_path(
             settings.RELE_GC_PROJECT_ID, topic)
         future = self._client.publish(topic_path, payload, **attrs)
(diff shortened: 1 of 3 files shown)


rele/client.py
```diff
@@ -49,8 +49,7 @@ def __init__(self):
 
     def publish(self, topic, data, blocking=False, **attrs):
         payload = JSONRenderer().render(data)
-        logger.info(f'Publishing to {topic}',
-                    extra={'pubsub_publisher_attrs': attrs})
+        self._save_log(attrs, topic)
```
tobami: A log is not "save", a message is "logged". To be consistent with how the subscriber logging it done, I would instead keep the logging inline and add a `_build_metrics` method for the publisher.

jjponz: Fixed. I add symmetry to publishers and subscribers

rele/client.py
```diff
@@ -50,7 +50,10 @@ def __init__(self):
     def publish(self, topic, data, blocking=False, **attrs):
         payload = JSONRenderer().render(data)
         logger.info(f'Publishing to {topic}',
-                    extra={'pubsub_publisher_attrs': attrs})
+                    extra={'pubsub_publisher_attrs': attrs,
+                           'metrics': self._build_metrics(topic),
```
tobami: This indentation should be same style as the way you've done in `_build_metrics`

rele/subscription.py
```diff
@@ -36,7 +36,8 @@ def __call__(self, message):
         db.close_old_connections()
 
         logger.debug(f'Start processing message for {self._subscription}',
-                     extra=self._build_metrics('received'))
+                     extra={'metrics':
+                                self._build_metrics('received')})
```
tobami: Again, let's make indentation consistent.

jjponz: Fix indentation

rele/client.py
jjponz: Fix indentation

rele/client.py
```diff
@@ -59,3 +61,12 @@ def publish(self, topic, data, blocking=False, **attrs):
 
         future.result(timeout=self.PUBLISH_TIMEOUT)
         return future
+
+    def _build_metrics(self, topic):
+        return{
```
andrewgy8:
```suggestion
        return {
```

rele/client.py
```diff
@@ -59,3 +61,12 @@ def publish(self, topic, data, blocking=False, **attrs):
...
+            'name': 'publications',
+            'data': {
+                'agent': settings.BASE_DIR.split('/')[-1],
```
andrewgy8: We should really think about making this configurable. We use this elsewhere, but this should not be the default behavior for widespread release.  In a later PR, instead this could be `settings.RELE_PROJECT_NAME`

Issue #8, opened 2019-05-13
Configurable serializer


tobami:
Publishing serialization is done by DRF's `rest_framework.renderers.JSONRenderer`. It should be configurable and default to `json.dumps`

andrewgy8:
Can we close this as it seems to be resolved with #82 and #83 ?

tobami:
Do we keep track elsewhere to remove the dependency?

andrewgy8:
I don't believe so. I recommend opening another issue called "Remove DRF dependency"

PR #25, opened 2019-06-03
Use drf jsonencoder, not renderer

tobami:
### :tophat: What?

Use json.dumps with drf's encoder, not the whole renderer.

### :thinking: Why?

It was simply incorrect. For example `.render(None)` was an empty string, which is not valid JSON.

### DIFF
--- a/rele/subscription.py
+++ b/rele/subscription.py
@@ -32,17 +32,14 @@ class Callback:
         self._suffix = suffix
 
     def __call__(self, message):
-        start_processing_time = time.time()
+        start_time = time.time()
         db.close_old_connections()
 
         logger.debug(f'Start processing message for {self._subscription}',
                      extra={
                          'metrics': self._build_metrics('received')
                      })
-        if message.data == b'':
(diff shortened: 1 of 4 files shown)


menecio:
Why not adding this to the existing PR? :thinking: You could already make the encoder class configurable and thru the settings

rele/subscription.py
```diff
@@ -32,42 +32,39 @@ def __init__(self, subscription, suffix=None):
...
 
         logger.debug(f'Start processing message for {self._subscription}',
                      extra={
                          'metrics': self._build_metrics('received')
                      })
-        if message.data == b'':
-            data = None
-        else:
-            data = json.loads(message.data.decode('utf-8'))
+        data = json.loads(message.data.decode('utf-8'))
```
andrewgy8: 👏

tobami:
@menecio sure, either in that PR or in a separate one implementing #8 that's what we'll do. We need this fix in the release though so we'll proceed with it for now

PR #26, opened 2019-06-04
Release rele 0.3.1

tobami:


andrewgy8:
Looks good. Perhaps we should add a `git tag` to the `make release` step.

Issue #7, opened 2019-05-13
Documentation


andrewgy8:
Sphinx? Read-the-docs? 

Any suggestions on documentation frameworks?

andrewgy8:
Im closing this as the issue seems solved with the RTD and sphinx integration. The only thing now is to add more docstrings to the project. Thanks @edgarlatorre !

PR #27, opened 2019-06-06
Initial sphinx documentation

edgarlatorre:
### :tophat: What?

Creates a initial structure for documentation using sphinx

* Add docs directory
* Index page and subscription module reading docstrings.
* Creates requirements/docs.txt with sphinx  as dependency
* Task to install dependencies to build the documentation page `make install-docs-requirements`
* Task to build documentation in html `make build-html-doc`

### Screenshot
Screenshot 2019-06-06 at 12 02 41

### :thinking: Why?

The idea is to write documentation using doctoring, so sphinx would be able to read it and create all the documentation in a website that can be read by GitHub pages or be hosted somewhere else. It's related to the issue #7


docs/conf.py
```diff
@@ -0,0 +1,59 @@
...
+import os
+import sys
+sys.path.insert(0, os.path.abspath('..'))
+
+import rele
+
+# -- Project information -----------------------------------------------------
+
+project = 'Rele'
+copyright = '2019, Mercadona'
```
andrewgy8:
```suggestion
copyright = '2019, Mercadona S.A.'
```

docs/conf.py
```diff
@@ -0,0 +1,59 @@
...
+#
+import os
+import sys
+sys.path.insert(0, os.path.abspath('..'))
+
+import rele
+
+# -- Project information -----------------------------------------------------
+
+project = 'Rele'
```
andrewgy8:
```suggestion
project = 'Relé'
```

Makefile
```diff
@@ -43,3 +43,9 @@ install-test-requirements: ## install requirements for testing
 
 install-deploy-requirements:  ## install requirements for deployment
 	pip install -r requirements/deploy.txt
+
+install-docs-requirements:  ## install requirements for documentation
```
andrewgy8: Would it be worth to have just one command `install-dev-requirements` which installs the `docs.txt` and `test.txt` requirements files?

edgarlatorre: Yeah it would be nice, but In my opinion it should be another PR.

docs/make.bat
```diff
@@ -0,0 +1,35 @@
+@ECHO OFF
```
jonasae: A batch file? Will this work on linux based systems?

edgarlatorre: Yes it is there in case a Windows user wants to generate the documentation. Linux based systems users would use the Makefile. This structure is generated like that by sphinx.

PR #41, opened 2019-06-11
Publish function docs

andrewgy8:
### :tophat: What?

Add some more docs to the project.
Also simplified the makefile

### :thinking: Why?

More docs the better

### DIFF
--- a/rele/publishing.py
+++ b/rele/publishing.py
@@ -19,7 +19,7 @@ def publish(topic, data, **kwargs):
             rele.publish(topic='lets-tell-everyone',
                        data={'foo': 'bar'},
                        myevent='arrival')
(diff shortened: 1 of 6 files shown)


andrewgy8:
Ok, and I added docs to the `worker` class...

rele/publishing.py
```diff
@@ -4,6 +4,28 @@
 
 
 def publish(topic, data, **kwargs):
+    """Shortcut method to publishing data to PubSub.
+
+    This is a shortcut method that instantiates the Publisher if not already
```
tobami: Alternative: "Publishes a message using a global Publisher to ensure long running processes don't leak memory"

Issue #31, opened 2019-06-07
Automatically add timestamp to published message


andrewgy8:
According to the [Google PubSub documentation, there is a `publishTime` attribute added to the message](https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage).

However, the caveat is that

> The time at which the message was published, populated by the server

IMHO, a UNIX timestamp could be added as a message attribute when the message is published from the client. This would prevent race conditions from the clients as the timestamp would be more accurate to the state of the system when publishing. 

We can following some standard examples of:

- [Dramatiq](https://github.com/Bogdanp/dramatiq/blob/master/dramatiq/message.py#L79)
- [Celery](https://github.com/celery/celery/blob/9f0a554dc2d28c630caf9d192873d040043b7346/celery/events/event.py#L29)
- [Kafka](https://kafka.apache.org/0100/javadoc/index.html?org/apache/kafka/connect/data/Timestamp.html)

And use `time.time()` safely.

PR #42, opened 2019-06-11
Add timestamp to published message

edgarlatorre:
### :tophat: What?

* Add optional parameter published_at  to publish function, it uses `time.time()` in case published_at is None.
* client.Publisher documentation page.

### Screenshot

Screenshot 2019-06-11 at 14 46 56

### :thinking: Why?

It can be useful to have the time that the message was published.

### DIFF
--- a/rele/client.py
+++ b/rele/client.py
@@ -49,7 +49,7 @@ class Publisher:
             self._client = pubsub_v1.PublisherClient(credentials=credentials)
 
     def publish(self, topic, data, blocking=False, **attrs):
-        """Publish data to Pub/Sub.
+        """Publishes data to Pub/Sub adding a timestamp `published_at` to the attrs.
 
         Usage::
 
--- a/rele/client.py
+++ b/rele/client.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import os
+import time
 from contextlib import suppress
 
 from django.conf import settings
@@ -48,6 +49,8 @@ class Publisher:
             self._client = pubsub_v1.PublisherClient(credentials=credentials)
 
     def publish(self, topic, data, blocking=False, **attrs):
+        attrs['published_at'] = str(time.time())
+
         logger.info(f'Publishing to {topic}',
                     extra={
                         'pubsub_publisher_attrs': attrs,
--- a/rele/client.py
+++ b/rele/client.py
@@ -48,7 +48,7 @@ class Publisher:
         else:
             self._client = pubsub_v1.PublisherClient(credentials=credentials)
 
-    def publish(self, topic, data, published_at=None, blocking=False, **attrs):
+    def publish(self, topic, data, blocking=False, **attrs):
         """Publish data to Pub/Sub.
 
         Usage::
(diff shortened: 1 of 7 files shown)


tests/test_client.py
```diff
@@ -1,43 +1,64 @@
...
+from unittest.mock import ANY, MagicMock, patch
 
 from google.cloud.pubsub_v1 import PublisherClient
 
 from rele import Publisher
 
 
 class TestPublisher:
+    @classmethod
+    def setup_method(cls):
```
tobami: We usually create a class-scoped fixture for this instead

edgarlatorre: Done.

andrewgy8:
Addresses #31

rele/client.py
```diff
@@ -49,6 +50,22 @@ def __init__(self):
                 credentials=settings.RELE_GC_CREDENTIALS)
 
     def publish(self, topic, data, blocking=False, **attrs):
+        """Publish data to Pub/Sub.
+
```
andrewgy8: I would suggest to add a sentence about the `published_at` time stamp attribute that we add.

edgarlatorre: @andrewgy8 I've changed the doc with your suggestion. Let me know if it's ok.

rele/client.py
```diff
@@ -49,6 +50,22 @@ def __init__(self):
...
+            publisher.publish('topic_name', {'foo': 'bar'})
+
+        :param topic: string topic to publish the data.
+        :param data: dict with the content of the message.
+        :param blocking: boolean, default False.
+        :param attrs: Extra parameters to be published.
+        :return: future
+        """
+
+        attrs['published_at'] = str(time.time())
```
tobami: Since casting to string is needed, shouldn't we do the opposite on the consuming side?

edgarlatorre: If every attribute in the message has to be a string I don't see the point of doing this. Also it should be implemented by the developer that is writing the subscription, isn't it?

andrewgy8: I think what @tobami is suggesting is that we deserialize the message data from bytes to a python dictionary. So we may want to serialize from a string timestamp to a python datetime object. Is that right?

tobami: Yep, same as we don't make the user decode and deserializer every payload, published_at has a very specific use case, so we'll always want it in the right type

edgarlatorre: Okay, just implemented it.

edgarlatorre: @tobami @andrewgy8 Can I merge it?

tests/test_client.py
```diff
@@ -1,46 +1,51 @@
...
-
-from . import settings
+from unittest.mock import ANY, patch
 
 
+@pytest.mark.usefixtures('publisher')
 class TestPublisher:
-
-    def test_returns_future_when_published_called(self):
+    @patch('time.time', return_value=1560244246.863829)
```
jonasae: We use to do this kind of patching (datetime, time) using a pytest fixture.

For instance:
```
@pytest.fixture
def published_at():
  return 1560244246.863829
@pytest.fixture
def time_mock(published_at):
  with patch('time.time') as mock:
    mock.return_value = published_at
    yield mock
```
Later on, as mock is not going to be referred to, but `published_at` will do:
```
@pytest.mark.usefixtures('publisher, time_mock')
def test(self, published_at, publisher):
...
assert_called_with(... published_at=published_at ...)
```
Furthermore, both fixtures will be very useful and reusable.

edgarlatorre: Cool, I'm gonna do that. Thanks!

Issue #38, opened 2019-06-10
Decouple django.conf.settings and rele settings


andrewgy8:
In a first step to decouple django and relé, we need agnostic place to put the relé settings. 

Right now, they are read directly from the project's `settings.py` file and then used like:
`from django.conf import settings`

We would also need to setup some default values, if applicable.

Suggestions:
- directly from env var
- another rele.settings file

To be honest, Im not entirely sure how we should handle this, so this is up for discussion. It will most likely require quite a bit of experimentation and trial and error.

PR #43, opened 2019-06-11
Explicitly pass publisher and subscriber configuration

tobami:
### :tophat: What?

Pass credentials as arguments for Publisher and Subscriber classes

### :thinking: Why?

So we can decouple from global settings


andrewgy8:
Addresses #38

rele/client.py
```diff
@@ -15,53 +15,51 @@
 
 class Subscriber:
 
-    def __init__(self):
+    def __init__(self, gc_project_id, credentials):
```
jonasae: Subscriber and Publisher has a pubsub client as colaborator.

Instead of passing credentials through the constructor, it would be cleaner to extract client's instantiation to an static function like `get_pubsub_client()` and either inject the client (oop) or instantiate in the constructor by doing `self._client = get_pubsub_client()`(pythonic).

PR #46, opened 2019-06-12
Add codecov

jonasae:
### :tophat: What?
Add codecov to the CI pipeline.

### :thinking: Why?
Why not?

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -1,5 +1,5 @@
 flake8>=2.1.0
-pytest
+pytest>=4.6.3
 coverage==4.4.1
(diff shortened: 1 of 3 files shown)


requirements/test.txt
```diff
@@ -1,4 +1,5 @@
-coverage==4.4.1
 flake8>=2.1.0
+pytest>=4.6.3
+coverage==4.4.1
+pytest-cov>=2.7.1
```
tobami: We don't need to pin so strictly, why not pytest 4.6.x and coverage 4.4.x?

PR #47, opened 2019-06-12
Subscriber prefix param

antoniobusrod:
### :tophat: What?

Use custom value as prefix to any subscription instead of static project name.

### :thinking: Why?

Small iteration in order to remove completely Django settings usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,9 @@ INSTALLED_APPS = [
 You'll also need to set up two variables with the Google Cloud credentials:
 `RELE_GC_CREDENTIALS` and `RELE_GC_PROJECT_ID`.
 
+You can use `RELE_SUB_PREFIX` to add a prefix to all your subscription names, for
+example, with your project name.
+
 NOTE: Ensure that [`CONN_MAX_AGE`](https://docs.djangoproject.com/en/2.2/ref/settings/#conn-max-age)
 is set to 0 in your worker. The Django default value is 0.
 
@@ -75,6 +78,17 @@ from rele import sub
 def sub_function(data, **kwargs):
       event = kwargs.get('myevent')
       print(f'I am a task doing stuff with an event: {event}')
+```
+
+#### Subscription `prefix`
+
+An optional prefix to the subscription name. Useful to namespace your subscription
+with your project name.
+
+```python
+@sub(topic='lets-tell-everyone', prefix='my-project')
+def purpose_1(data, **kwargs):
+     pass
 ```
 
 #### Subscription `suffix`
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -77,6 +80,17 @@ def sub_function(data, **kwargs):
       print(f'I am a task doing stuff with an event: {event}')
 ```
 
+#### Subscription `prefix`
```
andrewgy8: Shouldn't we be putting this sort of documentation in RTD?

tests/test_subscription.py
```diff
@@ -104,7 +104,7 @@ def test_log_when_execution_is_succesful(
 
     def test_log_does_not_ack_called_message_when_execution_fails(
             self, caplog, message_wrapper):
-        @sub(topic='some-cool-topic')
+        @sub(topic='some-cool-topic', prefix='rele')
```
andrewgy8: I would suggest adding a test that does not have the prefix `rele` since that is the default if no custom prefix is set.

rele/subscription.py
```diff
@@ -3,20 +3,29 @@
...
 from django import db
-from django.conf import settings
 
 logger = logging.getLogger(__name__)
 
 
 class Subscription:
 
-    def __init__(self, func, topic, suffix=None):
+    def __init__(self, func, topic, prefix=None, suffix=None):
```
andrewgy8: Could this be 
```suggestion
    def __init__(self, func, topic, prefix='', suffix=''):
```
instead? And then down below you don't have to do an `or`

README.md
antoniobusrod: Do you mean here?

https://github.com/mercadona/rele/pull/47/files#diff-ec9803a8e3588dfed144c3b60f46f1f1R119

andrewgy8: No, I mean we can put it in the index.rst page for now.  Later, we might want to have a configuration.rst page.

tests/test_subscription.py
antoniobusrod: 8e2acc5

rele/subscription.py
antoniobusrod: 5aa98ab

tests/test_worker.py
```diff
@@ -7,7 +7,7 @@
 from . import settings
 
 
-@sub(topic='some-cool-topic')
+@sub(topic='some-cool-topic', prefix='rele')
```
jonasae: Please remove this stub. Is duplicated and not used in worker's test context.

jonasae: Ok, now I see it's used not only here but in `test_subscription.py` as well.

Please move it to a `conftest.py` file in an upper level.

antoniobusrod: Because is not in other test modules, there is no reason to create `conftest.py`. Alternate approach should be to move it to the test function.

PR #49, opened 2019-06-12
Parametrize subscription ack deadline

### LINKED ISSUE #44: Parametrize subscription ack deadline
For any message for a topic (rele's subscription), use a custom ack deadline value instead of default 10 seconds using `ack_deadline_seconds` parameter in `SubscriberClient.create_subscription()` when creating a subscription.

It has been seen, that in some cases, there is a requirement to expend more than default 10 seconds to consume (late acknowledge) a message.

## Related

https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.subscriptions/modifyAckDeadline

daniel-ruiz:
### :tophat: What?

Add `DEFAULT_ACK_DEADLINE` variable to configure the default ack deadline (thus, overriding the default value set by Google). Plus, the `create_subscription` method in `Subscriber` class also accepts a custom ack deadline value.

### :thinking: Why?

The default value of 10 seconds set by Google may not be suitable in most of our use cases. Closes #44.

### DIFF
--- a/rele/client.py
+++ b/rele/client.py
@@ -24,7 +24,12 @@ class Subscriber:
         else:
             self._client = pubsub_v1.SubscriberClient(credentials=credentials)
 
-    def create_subscription(self, subscription, topic, ack_deadline_seconds=DEFAULT_ACK_DEADLINE):
+    def get_default_ack_deadline(self):
+        return int(os.environ.get('DEFAULT_ACK_DEADLINE', self.DEFAULT_ACK_DEADLINE))
+
+    def create_subscription(self, subscription, topic, ack_deadline_seconds=None):
+        ack_deadline_seconds = ack_deadline_seconds or self.get_default_ack_deadline()
+
         subscription_path = self._client.subscription_path(
             self._gc_project_id, subscription)
         topic_path = self._client.topic_path(self._gc_project_id, topic)
--- a/rele/client.py
+++ b/rele/client.py
@@ -15,6 +15,7 @@ USE_EMULATOR = True if os.environ.get('PUBSUB_EMULATOR_HOST') else False
 
 
 class Subscriber:
+    DEFAULT_ACK_DEADLINE = 60
 
     def __init__(self, gc_project_id, credentials):
         self._gc_project_id = gc_project_id
@@ -23,7 +24,7 @@ class Subscriber:
         else:
             self._client = pubsub_v1.SubscriberClient(credentials=credentials)
 
-    def create_subscription(self, subscription, topic, ack_deadline_seconds=60):
+    def create_subscription(self, subscription, topic, ack_deadline_seconds=DEFAULT_ACK_DEADLINE):
         subscription_path = self._client.subscription_path(
             self._gc_project_id, subscription)
         topic_path = self._client.topic_path(self._gc_project_id, topic)
--- a/rele/client.py
+++ b/rele/client.py
@@ -23,14 +23,16 @@ class Subscriber:
         else:
             self._client = pubsub_v1.SubscriberClient(credentials=credentials)
 
-    def create_subscription(self, subscription, topic):
+    def create_subscription(self, subscription, topic, ack_deadline_seconds=60):
         subscription_path = self._client.subscription_path(
             self._gc_project_id, subscription)
         topic_path = self._client.topic_path(self._gc_project_id, topic)
 
         with suppress(exceptions.AlreadyExists):
             self._client.create_subscription(
-                name=subscription_path, topic=topic_path)
+                name=subscription_path,
+                topic=topic_path,
+                ack_deadline_seconds=ack_deadline_seconds)
 
     def consume(self, subscription_name, callback):
         subscription_path = self._client.subscription_path(
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -1,7 +1,7 @@
-import pytest
 import concurrent
 from unittest.mock import ANY, patch
 
+import pytest
 from google.cloud.pubsub_v1 import SubscriberClient
 
 from rele.client import Subscriber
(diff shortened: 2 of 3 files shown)


rele/client.py
```diff
@@ -22,14 +23,21 @@ def __init__(self, gc_project_id, credentials):
         else:
             self._client = pubsub_v1.SubscriberClient(credentials=credentials)
 
-    def create_subscription(self, subscription, topic):
+    def get_default_ack_deadline(self):
+        return int(os.environ.get('DEFAULT_ACK_DEADLINE', self.DEFAULT_ACK_DEADLINE))
+
+    def create_subscription(self, subscription, topic, ack_deadline_seconds=None):
```
andrewgy8: What would be the use case for passing the `ack_deadline_seconds` into the `create_subscription` function if it can be set via envvar?

rele/client.py
```diff
@@ -14,6 +14,7 @@
 
 
 class Subscriber:
+    DEFAULT_ACK_DEADLINE = 60
```
andrewgy8: How about evaluating this conditional
```suggestion
    DEFAULT_ACK_DEADLINE = int(os.environ.get('DEFAULT_ACK_DEADLINE')) or 60
```
?

tests/test_client.py
```diff
@@ -44,3 +48,71 @@ def test_save_log_when_published_called(self, caplog):
...
+        return settings.RELE_GC_PROJECT_ID
+
+    @pytest.fixture()
+    def subscriber(self, project_id):
+        credentials = settings.RELE_GC_CREDENTIALS
+
+        return Subscriber(project_id, credentials)
+
+    @patch.object(SubscriberClient, 'create_subscription')
+    def test_create_subscription_without_ack_deadlines(
```
andrewgy8:
```suggestion
    def test_creates_subscription_with_60s_ack_deadline_when_none_provided(
```

tests/test_client.py
```diff
@@ -44,3 +48,71 @@ def test_save_log_when_published_called(self, caplog):
...
+
+        subscriber.create_subscription('test-topic',
+                                       f'{project_id}-test-topic')
+
+        _mocked_client.assert_called_once_with(ack_deadline_seconds=60,
+                                               name=expected_subscription,
+                                               topic=expected_topic)
+
+    @patch.object(SubscriberClient, 'create_subscription')
+    def test_create_subscription_with_ack_deadlines(
```
andrewgy8:
```suggestion
    def test_creates_subscription_with_configured_ack_deadline_when_provided(
```

rele/client.py
antoniobusrod: I would like to suggest to use the default 10 seconds as ack deadline

rele/client.py
daniel-ruiz: As far as we know, one subscriber instance is normally created to register several subscriptions into PubSub. Given this scenario, we thought there could be cases in which a single ack deadline couldn't be suitable to all the subscriptions inside a project. For those cases, the `ack_deadline_seconds` argument passed to `create_subscription` would be handy.

rele/client.py
daniel-ruiz: @andrewgy8 We thought about your suggestion at first place when implementing the feature. However, we wanted to go a step further and evaluate the `DEFAULT_ACK_DEADLINE` variable at the moment we actually use it. We thought this could add more flexibility to the settings and ease the unit testing as well. What do you think? Maybe we missed the point here.

About the default value. Since we complained a bit about the default ack deadline at the moment we discussed this feature, we opted to give our own default value. This is greater than the one Google sets and we also think it can be healthy to control our default value in order to provide more predictability (we don't know whether Google decides to change that default single-handedly at a certain moment).

tests/test_client.py
daniel-ruiz: Thanks for the suggestion, this better describes the test case. I only have one doubt: do you see useful to hardcode the expected value in the name of the test? That could be misleading in case this value changes.

rele/client.py
tobami: A good library also has reasonable defaults, i don't see a problem with using a bigger value than Google

rele/client.py
andrewgy8: If this is a feature we want to have we should add the `ack_deadline_seconds` argument to the `sub` decorator.  I suggest to break that feature into a separate issue, and keep this one using the env var to set the ack_deadline.

rele/client.py
andrewgy8: I agree that 60 seconds is fine.

tests/test_client.py
andrewgy8: I agree hard coding would not be wanted in this test. Instead, we are kind of documenting the default value in the test when no `ack_deadline` is set.

rele/client.py
daniel-ruiz: I agree. I created #54 to describe that feature as a separate one. We'll stick to the implementation of the global setting in this PR.

tests/test_client.py
```diff
@@ -1,6 +1,13 @@
...
 import concurrent
-from unittest.mock import ANY
+from unittest.mock import ANY, patch
+
+import pytest
+from google.cloud.pubsub_v1 import SubscriberClient
+
+from rele.client import Subscriber
+
+from . import settings
```
menecio: You should use `django.conf.settings` object

daniel-ruiz: As far as I know, we want to decouple the library from Django as far as possible. That said, and seeing that many other test modules apply the same pattern, I think we import settings that way to avoid importing `django`.

Does that make sense to you? I'm not 100% sure and want to better understand the case 😄

edgarlatorre: Actually I think we can remove that settings, we just use it to get the project_id and the credentials but now it's passed as parameter. I'll work to remove that.

PR #50, opened 2019-06-13
Pass the app_name to the publisher

andrewgy8:
### :tophat: What?

Passing `app_name` to the publisher

### :thinking: Why?

So we can remove the use of `django.conf.settings` in the client module.


andrewgy8:
Closing in deference to #55

PR #55, opened 2019-06-13
Middlewares

tobami:
### :tophat: What?

Move logging to a middleware.

### :thinking: Why?

So we can extend core functionality and decouple from Django


.editorconfig
```diff
@@ -7,7 +7,7 @@ indent_style = space
 
 [*.py]
 indent_size = 4
-max_line_length = 89
+max_line_length = 79
```
andrewgy8: Sneaky bugger

rele/middleware.py
```diff
@@ -0,0 +1,56 @@
+import importlib
+
+_middlewares = []
+
+
+def register_middleware(paths):
+    global _middlewares
```
andrewgy8: Perhaps this should be `_rele_middlewares`. That way there is even a smaller chance of conflicting with other packages. What do you think?

tobami: `_middlewares` is defined in the module. What happens outside it doesn't have any effect (unless we did a star import from somewhere, that is)

PR #61, opened 2019-06-14
Register contrib package in setup.py

daniel-ruiz:
### :tophat: What?

Register `rele.contrib` package in `setup.py`.

### :thinking: Why?

This way the package gets shipped with the library.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,6 +43,7 @@ setup(
     url='https://github.com/mercadona/rele',
     packages=[
         'rele',
+        'rele.contrib',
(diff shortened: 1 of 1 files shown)


andrewgy8:
🙀 

Great catch!

andrewgy8:
I wonder if there a way to include this in a test so it doesn't happen again.

daniel-ruiz:
> I wonder if there a way to include this in a test so it doesn't happen again.

Nice idea! Should we create an issue for that? 😄

PR #62, opened 2019-06-14
Fix Subscription import on runrele command.

edgarlatorre:
### :tophat: What?

We were calling rele.Subscription but we didn't import rele. Now I've changed it to import Subscription from rele.

### :thinking: Why?

`runrule` is not working.

### DIFF
--- a/rele/management/commands/runrele.py
+++ b/rele/management/commands/runrele.py
@@ -8,7 +8,7 @@ from django.conf import settings
 from django.core.management import BaseCommand
 from django.utils.module_loading import module_has_submodule
 
-from rele import config, Worker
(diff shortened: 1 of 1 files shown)


andrewgy8:
Good catch! This was fixed in #60 We actually moved that function out of the command. Now it should be easier to cover it with tests.

andrewgy8:
Do you want to add some more coverage for it in this PR?

daniel-ruiz:
> Good catch! This was fixed in #60 We actually moved that function out of the command. Now it should be easier to cover it with tests.

Due to this has been already solve, we close this PR. Thanks everyone for the feedback!

Issue #35, opened 2019-06-10
Filter by message attribute


andrewgy8:
We would like to have our subscribers filter messages from a particular topic based on message attribute(s). For instance, we have one publisher publishing to topic `foobar`. The message is published with either attribute `location='backyard'` or `location='frontyard'`. 

If we have two subscribers listening to the same topic, one should only act on `location='backyard'`. 

I don't want to publish to two separate topics as they are publishing the same data type, only the location attribute changes from message to message. 

Additionally, we would like to globally define the filter value in the app. If we have two workers running the same subscriber, one should be able to filter the messages only for `backyard` messages and the other for `frontyard`. 

This is a proposal implementation:

```
# Scenario A (Local usage)

# subs.py
@sub('update_location', filter_by={'location': 'backyard'})
def update_location_backyard():
    pass

@sub('update_location', filter_by={'location': 'frontyard'})
def update_location_frontyard():
    pass
```

```
# Scenario B (Global Setting)

# settings.py
RELE_FILTER_BY={
    'update_location': {
        'location': 'frontyard'
    }
}

# subs.py
@sub('update_location', filter_by=settings.RELE_FILTER_BY)
def update_location():
    pass
```

tobami:
In scenario B, does it make sense to set the `filter_by`? I understand "Global" as only needing to configure the global settings with a filter, no having to repeat myself in every `sub`.

For scenario A, a callable operating on the attributes would be more flexible, but more verbose:
```python
filter_by=labmda x: x['location'] == 'frontyard'
```

PR #66, opened 2019-06-14
Filter by message attributes

daniel-ruiz:
### :tophat: What?

Add a new parameter called `filter_by` to the `sub` decorator that filters the messages to be processed. The filtering is based on the message attributes.

### :thinking: Why?

There are messages sent to a subscription that shouldn't be processed.

### :link: Related issue

Refs #35

### DIFF
--- a/tests/test_subscription.py
+++ b/tests/test_subscription.py
@@ -60,14 +60,17 @@ class TestSubscription:
         assert log2.message == 'I am a task doing stuff with ID 123 (es)'
 
     def test_sub_executes_when_message_attributes_match_criteria(self, caplog):
-        sub_process_landscape_photos({'name': 'my_new_photo.jpeg'}, type='landscape')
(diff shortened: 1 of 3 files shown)


tests/test_subscription.py
```diff
@@ -30,6 +30,16 @@ def sub_published_time_type(data, **kwargs):
...
 
 
+def landscape_filter(**kwargs):
+    return kwargs.get('type') == 'landscape'
+
+
+@sub(topic='photo-updated', filter_by=landscape_filter)
+def sub_process_landscape_photos(data, **kwargs):
+    photo_type = kwargs.get('type')
+    logger.info(f'Received a photo of type {photo_type}')
```
antoniobusrod: I'd like to suggest to return a value to avoid inspecting log messages

edgarlatorre: Much better, thanks!

PR #72, opened 2019-06-17
Filter by message attributes

### LINKED ISSUE #35: Filter by message attribute

edgarlatorre:
### :tophat: What?

Use default filter_by defined on settings.

`
{
  'FILTER_SUBS_BY': lambda attrs: attrs.get('type') == 'foo'
}
`

### Screenshot
Documentation:

Screenshot 2019-06-18 at 10 01 44

### :thinking: Why?

There's cases where we want to have a default filter for all subscriptions.

### :link: Related issue

Fix #35

### DIFF
--- a/rele/config.py
+++ b/rele/config.py
@@ -22,7 +22,7 @@ def setup(setting):
     return config
 
 
-def load_subscriptions_from_paths(sub_module_paths, sub_prefix=None):
+def load_subscriptions_from_paths(sub_module_paths, sub_prefix=None, filter_by=None):
     subscriptions = []
     for sub_module_path in sub_module_paths:
         sub_module = importlib.import_module(sub_module_path)
@@ -31,5 +31,9 @@ def load_subscriptions_from_paths(sub_module_paths, sub_prefix=None):
             if isinstance(attribute, Subscription):
                 if sub_prefix and not attribute.prefix:
                     attribute.set_prefix(sub_prefix)
+
+                if filter_by and not attribute.filter_by:
+                    attribute.set_filter_by(filter_by)
+
                 subscriptions.append(attribute)
     return subscriptions
--- a/rele/config.py
+++ b/rele/config.py
@@ -22,7 +22,9 @@ def setup(setting):
     return config
 
 
-def load_subscriptions_from_paths(sub_module_paths, sub_prefix=None, filter_by=None):
+def load_subscriptions_from_paths(
+        sub_module_paths, sub_prefix=None, filter_by=None):
+
     subscriptions = []
     for sub_module_path in sub_module_paths:
         sub_module = importlib.import_module(sub_module_path)
--- a/rele/subscription.py
+++ b/rele/subscription.py
@@ -41,13 +41,16 @@ class Subscription:
         if 'published_at' in kwargs:
             kwargs['published_at'] = float(kwargs['published_at'])
 
-        if self._filter_by and not self._filter_by(kwargs):
+        if self._filter_returns_false(kwargs):
             return
 
         return self._func(data, **kwargs)
 
     def __str__(self):
         return f'{self.name} - {self._func.__name__}'
+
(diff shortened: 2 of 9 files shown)


rele/config.py
```diff
@@ -31,5 +33,9 @@ def load_subscriptions_from_paths(sub_module_paths, sub_prefix=None):
             if isinstance(attribute, Subscription):
                 if sub_prefix and not attribute.prefix:
                     attribute.set_prefix(sub_prefix)
+
+                if filter_by and not attribute.filter_by:
```
jonasae: Setting prefix and filter must be handled by `Subscription`.
Therefore I suggest moving the conditions inside `set_prefix` and `set_filter_by`.

edgarlatorre: This is to set a default filter from the settings, I'm not sure it should be handled by the Subscription.

rele/subscription.py
```diff
@@ -30,14 +30,21 @@ def prefix(self):
...
+    def set_filter_by(self, filter_by):
+        self._filter_by = filter_by
+
     def __call__(self, data, **kwargs):
         if 'published_at' in kwargs:
             kwargs['published_at'] = float(kwargs['published_at'])
 
-        if self._filter_by(**kwargs):
-            return self._func(data, **kwargs)
+        if self._filter_by and not self._filter_by(**kwargs):
```
jonasae: Sorry, I don't understand what to expect from `self._filter_by(**kwargs)`

What do you think about extracting the condition to a meaningful method such `self._guard_against_processing_unfiltered_messages`?

tests/settings.py
```diff
@@ -35,6 +35,7 @@
         'rele.contrib.LoggingMiddleware',
         'rele.contrib.DjangoDBMiddleware',
     ],
+    'FILTER_SUBS_BY': lambda attrs: attrs.get('type') == 'foo'
```
daniel-ruiz: Is this properly working? I see in the examples the callback is declared with `attrs` being a keyword parameter: 
```python
    def subscriptions(self):
        return load_subscriptions_from_paths(
            ['tests.test_config'],
            sub_prefix='test',
            filter_by=lambda **attrs: attrs.get('lang') == 'en')
```

Is this something that can make this global filter fail? 🤔

edgarlatorre: We are not using the settings.py so I'm gonna remove the filter from there. Also I'll change the filter_by in the subscription fixture.

rele/subscription.py
edgarlatorre: Yeah, good idea to extract it, I'm going to do that.

rele/config.py
tobami: This is not ideal but the only way to avoid it would be to implement global settings so that the `sub` decorator can read from them when instantiating the `Subscription` instances. I'd say let's handle that issue separately, since this PR uses the existing mechanism correctly.

PR #73, opened 2019-06-18
Add showsubscriptions command

### LINKED ISSUE #15: Self documentation of topic-subscribers in a project
As the number of topics/subscribers increases in a project, it becomes harder and harder to figure out where/what is being subscribed too. 

It would be super cool to have a command like `rele document` which would output to the console a list of the topics and their subscribers with the sub name. That way you could get a quick glance at what the project's subscribers are doing.  

Something like:

Topic | Subscriber(s)  | Sub
------------ | ------------- | ----------
do-something | lets-do-something | my_lets_do_something_subscriber
write-something-else | lets-write-something-else | my_lets_write_something_else_subscriber
write-something-else | lets-write-something-else-suffix | other_lets_write_something_else_subscriber

daniel-ruiz:
### :tophat: What?

Add a `showsubscriptions` Django command that outputs the information of the PubSub subscriptions registered using Relé.

rele_document_command

### :thinking: Why?

It is convenient to have such command that links callback function names with their PubSub subscription names and topics. Especially in large projects that have many subscriptions.

### :link: Related issue

Closes #15


rele/management/commands/document.py
```diff
@@ -0,0 +1,30 @@
+from django.apps import apps
+from django.core.management import BaseCommand
+from django.utils.module_loading import module_has_submodule
+from tabulate import tabulate
+
+from rele.config import load_subscriptions_from_paths
+
+
+def discover_subs_modules():
```
edgarlatorre: I think we have a very similar function in the file runrele.py `_discover_subs_modules`. What do you think about to move this function somewhere and use the same function in both cases?

tests/commands/test_document.py
```diff
@@ -0,0 +1,64 @@
...
+            ('published-time-type',
+             'published-time-type',
+             'sub_published_time_type'),
+            ('some-cool-topic', 'rele-some-cool-topic', 'sub_stub'),
+            ('some-fancy-topic', 'some-fancy-topic', 'sub_fancy_stub'),
+        ]
+
+        call_command('document')
+
+        mock_tabulate.assert_called_once_with(expected_subscription_data,
```
tobami: Instead of mocking you can actually test that the subscription names are in the std output

rele/management/commands/document.py
```diff
@@ -0,0 +1,30 @@
+from django.apps import apps
```
tobami: Management command names are in a global namespace. `python manage.py document` is too generic in the django-wide context. It could be called something like `showsubscriptions`, which also matches the pattern other commands follow, like `showmigrations`

andrewgy8:
@daniel-ruiz This looks awesome!

rele/management/commands/document.py
daniel-ruiz: I agree with you @edgarlatorre . I was aware this function was duplicated but I didn't know how to remove that duplication. What if we move this function to `rele.config` module? Do you have any suggestions? 🤔

andrewgy8: We could create a `ReleBaseCommand` and use it in both commands. That would be my suggestion.

edgarlatorre: My first thought would move the function to `rele.config`.  I wouldn't create a base command for that as `discover subscription` is not responsibility of a command.

andrewgy8: The issue with moving it to `rele.config` is that it has django django dependencies. And we don't want to pollute what will eventually be rele standalone with more django dependencies.  

The other option I could see is moving it to a `rele.management.commands.discover` module.

edgarlatorre: What about to have something like a loader.py module? I believe it'll be necessary to use this function for a standalone version.

tobami: This could be moved to `rele.contrib.django`

daniel-ruiz: I like what @andrewgy8 says about keeping Django dependencies as isolated as possible and agree with @edgarlatorre in that it makes little sense to create a `BaseCommand` subclass only for adding this function. 

I'll try to move the function to `rele.management.commands.discover` module so that the relative import statement is short enough. Let's see how it looks 😄

rele/management/commands/showsubscriptions.py
```diff
@@ -0,0 +1,20 @@
+from django.core.management import BaseCommand
+from tabulate import tabulate
+
+from rele.config import load_subscriptions_from_paths
+
+from rele.management.discover import discover_subs_modules
+
+
+class Command(BaseCommand):
+    help = 'List information about the subscriptions registered using Relé.'
```
tobami: "... about Pub/Sub subscriptions ..."

rele/management/discover.py
```diff
@@ -0,0 +1,18 @@
+import logging
+
+from django.apps import apps
+from django.utils.module_loading import module_has_submodule
+
+logger = logging.getLogger(__name__)
+
+
+def discover_subs_modules():
```
andrewgy8: How about naming this `_subs_modules`?

daniel-ruiz: And using this function like `discover._subs_modules`?

daniel-ruiz: But accessing to an external function whose name starts with `_` isn't a good practice, is it? 😕

Issue #65, opened 2019-06-14
Style guide


andrewgy8:
We currently have flake8 in our linting command.  We do not have isort in the project style guidelines. 

Shall we add it? 

Or try out something new like [black](https://github.com/python/black)?

daniel-ruiz:
I have no problem with undertaking either action. To me, black is something I wanted to try out. I see this project could be a good experimentation field for this. After all, it is widely used in other Python open-source projects.

tobami:
I'd add isort, but not `black` yet. Not unless we have formatting problems.

daniel-ruiz:
Why? I never used `black`, so I may have missed something. Do you see any drawbacks of using it at this point?

tobami:
I asked Andrew what he meant and he didn't mean to use `black` for the autoformatting capabilities but as a substitute for flake8 for the reason that it has checks beyond what flake8 offers. In that case I'm OK with trying it out.

edgarlatorre:
After quick read in the black documentation, IMHO I think it'd be kind of redundant. The differences from fake8 is that black would suggest changes to follow the defined style and for me it would work super well in a private project where you can have an agreement with the developer to use black integrated with you editor. For a open source project I wouldn't mind to have different styles if it does not have any pep8 offence. Also if we are not using auto format, it's kind of annoying to check the diffs with the suggestions.

PR #77, opened 2019-06-20
Try out black

andrewgy8:
### :tophat: What?

Apply black linting. It seems to be mostly single --> double quotes. 

### :thinking: Why?

Apply an opinionated formatter as discussed in #65 

### :link: Related issue

#65

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,17 @@
 dist: xenial
-language: python
+sudo: false
+notifications:
+  email: false
 
 services:
   - postgresql
-
 addons:
   postgresql: "9.6"
 
+language: python
 python:
   - "3.6"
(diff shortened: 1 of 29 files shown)


edgarlatorre:
Are we supposed to run black before every PR? If it is yes I think we should specify it somewhere. If not I have the impression that we are going to have a lot of necessary conflicts because someone forgot to run black.

andrewgy8:
It'll get checked in Travis to make sure it passes the linting stage. I don't foresee any conflict issues, as the tool behaves quite similarly to other linting tools we use.  You could run black before every PR in the case that your code is not compliant with this linting tools configuration. Very much like I use `isort` in other projects.

PR #82, opened 2019-07-05
Json encoder class is configurable in the Publisher client Class

andrewgy8:
### :tophat: What?

Allow the Publisher class to have a configurable json encoder.

### :thinking: Why?

One small step for making the library django dependency free. 

### :link: Related issue

Addresses #8

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -1,3 +1,4 @@
+import json
 import os
 import concurrent
 from unittest.mock import ANY, patch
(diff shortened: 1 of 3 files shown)


tests/test_client.py
```diff
@@ -44,6 +45,19 @@ def test_publish_sets_published_at(self, published_at, publisher):
             ANY, b'{"foo": "bar"}', published_at=str(published_at)
         )
 
+    def test_publishes_data_with_custom_encoder(self, publisher):
+        import decimal
+
+        class DecimalEncoder(json.JSONEncoder):
```
tobami: Nice!

andrewgy8:
Yeah! I forgot about that in the PR description. 

But the next step will be to add a JSONEncoder field to the config object. That way we can add a default JSON encoder in the settings. This will eventually remove the DRF dependency from rele.

tobami:
Well the first step can be adding the config while still defaulting to drf's encoder. That way projects can start to adapt at their own pace.

PR #84, opened 2019-07-11
ENCODER settings config

andrewgy8:
### :tophat: What?

Add the `ENCODER` setting to the config object.

### :thinking: Why?

It allows the global publisher to be initialized with the desired encoder. 

### :link: Related issue

Addresses #8

### DIFF
--- a/docs/api/settings.rst
+++ b/docs/api/settings.rst
@@ -27,7 +27,7 @@ Example::
         ],
         'SUB_PREFIX': 'mysubprefix',
         'APP_NAME': 'myappname',
-        'ENCODER': encoders.CustomEncoderClass,
+        'ENCODER': 'rest_framework.utils.encoders.JSONEncoder',
     }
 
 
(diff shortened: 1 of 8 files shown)


docs/api/settings.rst
```diff
@@ -86,3 +87,17 @@ The application name.
...
+
+**Optional**
+
+Default: `rest_framework.utils.encoders.JSONEncoder <https://github.com/encode/django-rest-framework/blob/master/rest_framework/utils/encoders.py#L17>`_
+
+`Encoder class <https://docs.python.org/3/library/json.html#json.JSONEncoder>`_ to use for
+serializing your Python data structure to a json object when publishing.
+
+NOTE: The default encoder class is subject to change in an upcoming release. It is
+strongly advised that you use this setting explicitly.
```
tobami: We do not really need to set it explicitly *unless* you depend on a superset of the standard JSON encoder.

rele/config.py
```diff
@@ -22,6 +23,7 @@ def __init__(self, setting):
         self.app_name = setting.get("APP_NAME")
         self.sub_prefix = setting.get("SUB_PREFIX")
         self.middleware = setting.get("MIDDLEWARE", default_middleware)
+        self.encoder = setting.get("ENCODER", DEFAULT_ENCODER)
```
tobami: This should work like the middleware setting: a dotted path lazy import. Otherwise you'll have to import the class into your settings file.

docs/api/settings.rst
andrewgy8: Right. In most cases we will fall back to the optional default. But, as of now we are guaranteeing that `serializer.data` can be serialized via a superset of JSONEncoder. And as far as I know, we won't in the future. If set, it would allow breaking changes, to be less severe.  But if you think its too much overhead, then Ill change it.

andrewgy8: Unless of course, we want to set it by default to `json.encoder.JSONEncoder`. It would be the eventual step that needs to be taken. I guess do we do it now, or later?

rele/client.py
```diff
@@ -68,9 +68,7 @@ class Publisher:
     :param timeout: integer, default 3.0 seconds.
     """
 
-    def __init__(
-        self, gc_project_id, credentials, encoder=encoders.JSONEncoder, timeout=3.0
-    ):
+    def __init__(self, gc_project_id, credentials, encoder, timeout=3.0):
```
tobami: `encoder` could default to the std lib json one

PR #87, opened 2019-07-26
Move default ack deadline to the rele config dict

### LINKED ISSUE #83: DEFAULT_ACK_DEADLINE should be moved to the Config Object
In the publisher class, there is the use of DEFAULT_ACK_DEADLINE. 

I propose that this should be moved to the config object so it can be set via the settings dictionary.

andrewgy8:
### :tophat: What?

In attempt to consolidate all the rele settings out there into one rele dict setting object, here is the moving of the configuration for the default ack_deadline. 

### :link: Related issue

Fix #83

### DIFF
--- a/rele/management/commands/runrele.py
+++ b/rele/management/commands/runrele.py
@@ -5,8 +5,8 @@ import time
 from django.conf import settings
 from django.core.management import BaseCommand
 
-from rele import Worker
 import rele
+from rele import Worker
 from rele.config import Config
 
 from rele.management.discover import discover_subs_modules
--- a/rele/management/commands/runrele.py
+++ b/rele/management/commands/runrele.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import signal
 import time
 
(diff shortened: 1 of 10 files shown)


andrewgy8:
I left this as a draft for the time being. If we agree to move on, Ill add docs to this a submit a PR.

rele/management/commands/runrele.py
```diff
@@ -22,7 +23,12 @@ def handle(self, *args, **options):
         for sub in subs:
             self.stdout.write(f"  {sub}")
         worker = Worker(
-            subs, settings.RELE["GC_PROJECT_ID"], settings.RELE["GC_CREDENTIALS"]
+            subs,
+            settings.RELE["GC_PROJECT_ID"],
+            settings.RELE["GC_CREDENTIALS"],
+            settings.RELE.get(
+                "DEFAULT_ACK_DEADLINE", os.environ.get("DEFAULT_ACK_DEADLINE", 60)
```
tobami: This belongs in `config.py`

PR #104, opened 2019-09-28
Prepare 0.7.0 release

andrewgy8:
### :tophat: What?

Prepare 0.7.0 release

PR #108, opened 2019-10-18
Split django specific docs and standalone docs

andrewgy8:
### :tophat: What?

Add Django and standalone documentation.

Screen Shot 2019-10-18 at 5 50 40 PM
Screen Shot 2019-10-18 at 5 50 47 PM
Screen Shot 2019-10-18 at 5 50 56 PM

### :thinking: Why?

Preparing for the 0.7.0 release which remove django as a requirement

### :link: Related issue

#104

### DIFF
--- a/rele/contrib/logging_middleware.py
+++ b/rele/contrib/logging_middleware.py
@@ -46,7 +46,7 @@ class LoggingMiddleware(BaseMiddleware):
         )
 
     def pre_process_message(self, subscription, message):
-        self._logger.debug(
+        self._logger.info(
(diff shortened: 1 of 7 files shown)


rele/contrib/logging_middleware.py
```diff
@@ -46,7 +46,7 @@ def pre_publish(self, topic, data, attrs):
         )
 
     def pre_process_message(self, subscription, message):
-        self._logger.debug(
+        self._logger.info(
```
tobami: Was this change on purpose?

PR #109, opened 2019-10-21
Update readme to include the quickstart in the guide

andrewgy8:
### :tophat: What?

Update the readme

### :thinking: Why?

We should point to the docs for a quickstart. And also include the install directions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,7 @@
 
 <p align="center">
     <strong>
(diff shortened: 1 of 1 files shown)


andrewgy8:
This is what it will look like https://github.com/mercadona/rele/blob/c8a4b1a63a6c8e0d0de646753ef70675ce23056f/README.md

Issue #114, opened 2019-11-01
Simplify initializing and running the Worker


andrewgy8:
One consistent peice of feedback that I have received is simplifying the Worker class. Right now, you must Initialize with the `subs`, each individual `config` attribute, run `setup`, run `start`, and then `sleep`.

Like this:

```
worker = Worker(
    [photo_uploaded],
    config.gc_project_id,
    config.credentials,
    config.ack_deadline,
)
worker.setup()
worker.start()
sleep(120)
```

I propose we simplify the API to run a worker to look something like this instead:

```
worker = Worker([photo_uploaded], config)
worker.run(sleep=120)
```

`run` would call both `setup` and `start`, and we could add the standard `sleep` method. In addition we consolidate the configuration into one attribute. 

This would be backwards compatible since we will be creating a new method. And the change to Worker initialization could also fall back to the declared attributes if defined. Otherwise, use the `config` object.

tobami:
Sounds like a great plan to me!

hobochild:
Not sure this is convenient or possible but If you already have an api for setting up `rele` singleton with the required config, can't the worker use that api too? Then the setup for publishing and the Worker would be the same? 

eg
```
# config.setup should always be called first.
rele.config.setup({
    'GC_CREDENTIALS': service_account.Credentials.from_service_account_file(
        GOOGLE_APPLICATION_CREDENTIALS
    ),
    'GC_PROJECT_ID': GOOGLE_PROJECT_ID,
})

worker = rele.Worker(
    [photo_uploaded],
)
worker.setup()
worker.start()
sleep(120)
```

tobami:
That's certainly an option! We could make the config parameters optional in case the config was already initialized

andrewgy8:
Released in #131

PR #118, opened 2019-11-09
Worker run method

andrewgy8:
### :tophat: What?

Add a shortcut method for calling setup, start, and sleep. 

### :thinking: Why?

Make it easier to start a worker.

### :link: Related issue

#114

### DIFF
--- a/rele/worker.py
+++ b/rele/worker.py
@@ -50,6 +50,10 @@ class Worker:
             )
         run_middleware_hook("post_worker_start")
 
+    def run(self):
+        self.setup()
+        self.start()
+
     def stop(self, signal=None, frame=None):
         """Manage the shutdown process of the worker.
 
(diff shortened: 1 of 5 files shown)


andrewgy8:
Yeah I will do that. I just didnt want the diff to get all cluttered.

rele/worker.py
```diff
@@ -48,6 +49,15 @@ def start(self):
                 )
             )
 
+    def run(self, wait=60):
```
tobami: A `run` that ends after a given number of seconds does not make things much easier. I would instead implement a `run_forever(sleep_interval=2)`, for example.

hobochild: Yea this would be great, maybe something like: 

```

andrewgy8: By popular demand, I added the `run_forever` method. 👍

PR #119, opened 2019-11-17
New worker initialization format

EmilioCarrion:
### :tophat: What?

Simplified worker initialization.
This allows you to create workers with the rele config directly.
This keeps compatibility with the old initialization format.

### :thinking: Why?

Currently you must pass the config parameters in the worker creation.

### :link: Related issue

This is part of this enhancement issue: https://github.com/mercadona/rele/issues/114

### DIFF
--- a/docs/guides/basics.rst
+++ b/docs/guides/basics.rst
@@ -128,12 +128,7 @@ and will begin to pull the messages from the topic.
     from subs import photo_uploaded
 
     if __name__ == '__main__':
-        worker = Worker(
-            [photo_uploaded],
-            config.gc_project_id,
-            config.credentials,
-            config.ack_deadline,
-        )
+        worker = Worker([photo_uploaded], config)
         worker.run_forever()
 
 Once the sub and worker are created, we can start our worker by running ``python worker.py``.
--- a/rele/management/commands/runrele.py
+++ b/rele/management/commands/runrele.py
@@ -30,12 +30,7 @@ class Command(BaseCommand):
         self.stdout.write(f"Configuring worker with {len(subs)} " f"subscription(s)...")
         for sub in subs:
             self.stdout.write(f"  {sub}")
-        worker = Worker(
-            subs,
-            self.config.gc_project_id,
-            self.config.credentials,
-            self.config.ack_deadline,
-        )
+        worker = Worker(subs, self.config)
 
         signal.signal(signal.SIGINT, signal.SIG_IGN)
         signal.signal(signal.SIGTERM, worker.stop)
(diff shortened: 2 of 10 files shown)


rele/worker.py
```diff
@@ -17,7 +18,18 @@ class Worker:
     :param subscriptions: list :class:`~rele.subscription.Subscription`
     """
 
-    def __init__(self, subscriptions, gc_project_id, credentials, default_ack_deadline):
+    def __init__(self, subscriptions, *args):
+        if args[0] == rele.config:
```
andrewgy8: Im not sure how this is passing since `rele.config` is a module and `args[0]` is an instance of a `Config` class. I wonder if it has to do with what [the test](https://github.com/mercadona/rele/pull/119/files#diff-3ba621217225120eff2c061666b8043cR68) is asserting. 

Perhaps we should do:

```suggestion
        if isinstance(args[0], Config):
```

And in addition add some explicit assertions in the test, like:

```
assert worker.credentials == ...

```

rele/worker.py
```diff
@@ -1,6 +1,7 @@
 import logging
 import sys
 
+import rele
```
andrewgy8:
```suggestion
from .config import Config
```

rele/management/commands/runrele.py
```diff
@@ -31,12 +31,7 @@ def handle(self, *args, **options):
         self.stdout.write(f"Configuring worker with {len(subs)} " f"subscription(s)...")
         for sub in subs:
             self.stdout.write(f"  {sub}")
-        worker = Worker(
-            subs,
-            self.config.gc_project_id,
-            self.config.credentials,
-            self.config.ack_deadline,
-        )
+        worker = Worker(subs, self.config)
```
andrewgy8: 👏

docs/guides/basics.rst
```diff
@@ -128,12 +128,7 @@ and will begin to pull the messages from the topic.
     from subs import photo_uploaded
 
     if __name__ == '__main__':
-        worker = Worker(
-            [photo_uploaded],
-            config.gc_project_id,
-            config.credentials,
-            config.ack_deadline,
-        )
```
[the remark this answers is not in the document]
tobami: It's simpler but I'm trying to decide whether passing a single "config object" argument is better than explicit arguments. It could instead be that the broker/backend could be injected as an argument, itself initialized with the proper configuration.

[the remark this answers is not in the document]
andrewgy8: You could also inject the entire config object into the Worker which would have the backend initialized (like here).

PR #121, opened 2019-11-19
Add missing worker middleware hooks

hobochild:
### :tophat: What?

Added missing calls to documented hooks

### :thinking: Why?

I was checking out all the middleware hooks and noticed these two were not called. I'm not actually using these hooks so I'm not sure they are necessary but they are mentioned in the docs and implemented in the MiddlwareBase class.


andrewgy8:
Thanks for the contrib @hobochild ! 👏

Issue #120, opened 2019-11-19
Add an API to allow passing objects to middleware


hobochild:
I'm adding rele to flask app, and the subscription callbacks need the flask app_context. I've done this by way of middleware: 

eg:
```
class FlaskMiddleware(BaseMiddleware):
    def pre_process_message(self, subscription, message):
        from server import app
        self.ctx = app.app_context()
        self.ctx.push()

    def post_process_message(self):
        self.ctx.pop()
```

But to make this reusable across our services and other flask apps, I'd need a way to add arbitrary data to the config that is passed to middleware.setup method or have an easy way to call custom middleware functions.

eg:
```
class FlaskMiddleware(BaseMiddleware):
    def setup(self, config):
        self.app = config["FLASK_APP"]

    def pre_process_message(self, subscription, message):
        self.ctx = self.app.app_context()
        self.ctx.push()

    def post_process_message(self):
        self.ctx.pop()
```

andrewgy8:
Flask integration sounds like a great addition and natural progression. 👏 
 
If I understand correctly though, the real problem to solve here would be adding extra data to the `Config` class. In this case it would be the Flask `App`. 

One solution that came to mind, and I'm not sure how elegant it is, would be to add `config.extras` which would be a way to add whatever you want to the configuration. I'm not super happy with that solution, as the config could get very messy, unorganized, and unpredictable.

Another to solve this issue would be to add `args/kwargs` to the `config.setup` method, passing them through to the `register_middleware` and then to the `BaseMiddleware.setup`. That would allow anyone calling setup to configure their app accordingly. Better, but still a bit convoluted.

I was looking at how Celery does it, and [to be honest its not pretty](https://flask.palletsprojects.com/en/1.1.x/patterns/celery/#configure). 

These are some thoughts on this, and perhaps someone else has a better solution.

tobami:
`middleware.setup()` gets the config object as an argument, so what you describe can be implemented exactly that way.

hobochild:
@andrewgy8 yea that's exactly the usecase. I think passing kwargs + config to setup function is probably the cleanest.

@tobami, unfortunately not as `Config` is strict with what takes from settings https://github.com/mercadona/rele/blob/master/rele/config.py#L10

andrewgy8:
Closing as it was addressed in #123 and will be in the next release.

PR #123, opened 2019-11-21
Add kwargs to setup method passed through to middleware

andrewgy8:
### :tophat: What?

Add kwargs to the setup method and pass them through to the BaseMiddleware setup function

### :thinking: Why?

Allows users to configure custom configuration for Middleware.

### :link: Related issue

Possible fix for #120

### DIFF
--- a/rele/middleware.py
+++ b/rele/middleware.py
@@ -6,7 +6,7 @@ _middlewares = []
 default_middleware = ["rele.contrib.LoggingMiddleware"]
 
 
-def register_middleware(config):
+def register_middleware(config, **kwargs):
     paths = config.middleware
     global _middlewares
     _middlewares = []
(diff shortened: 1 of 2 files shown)


rele/middleware.py
```diff
@@ -6,7 +6,7 @@
 default_middleware = ["rele.contrib.LoggingMiddleware"]
 
 
-def register_middleware(config):
+def register_middleware(config, **kwargs):
```
tobami: I don't see how this is going to solve the issue. `register_middleware` is called from `config.setup`, so it won't get any parameters anyway. We would need an extra mechanism for this.

rele/middleware.py
```diff
@@ -15,7 +15,7 @@ def register_middleware(config):
         module_path = ".".join(module_parts)
         module = importlib.import_module(module_path)
         middleware = getattr(module, middleware_class)()
-        middleware.setup(config)
```
tobami: Seeing this again, we probably could do without `middleware.setup`. All that could be done in the init

rele/middleware.py
tobami: OK, let's do it like this

hobochild:
Will give this a spin on Monday!

andrewgy8:
Great! Ill try to do a release ASAP as soon as I merge #118

PR #127, opened 2019-11-26
Rele flask

hobochild:
### :tophat: What?

Adds middleware for easy integration with Flask and fixes issue with logging Middleware not expecting kwargs.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -3,10 +3,11 @@ Thank you to all the contributors to Relé!
 * Andrew Graham-Yooll (@andrewgy8)
 * Miquel Torres Barcelo (@tobami)
 * Edgar Latorre (@edgarlatorre)
-* Lluis Guirado (@LluisGuiVi) 
-* Daniel Ruiz (@dani-ruiz) 
+* Lluis Guirado (@LluisGuiVi)
+* Daniel Ruiz (@dani-ruiz)
 * Juanjo Ponz (@jjponz)
 * Jose Antonio Navarro (@jonasae)
 * Antonio Bustos Rodriguez (@antoniobusrod)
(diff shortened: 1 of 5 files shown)


hobochild:
For some reason `TestPublish.test_creates_global_publisher_when_published_called` test case was failing I didn't touch it so assume it was something to do with my local setup ? will wait for the CI to tell me :)

hobochild:
Not sure whats up with the tests - but it appears that calling `rele.config.setup()` in the new middleware test causes the `test_creates_global_publisher_when_published_called` to fail. So must have some shared state between the tests somehow. :/

andrewgy8:
This looks great! After playing around on your fork, the issue was indeed the global publisher. And this snippet should solve the issue:

```
import pytest

import rele
from unittest.mock import patch


class TestMiddleware:

    @pytest.fixture
    def mock_init_global_publisher(self):
        with patch("rele.config.init_global_publisher", autospec=True) as m:
            yield m

    @pytest.mark.usefixtures('mock_init_global_publisher')
    @patch("rele.contrib.FlaskMiddleware.setup", autospec=True)
    def test_setup_fn_is_called_with_kwargs(
        self, mock_middleware_setup, project_id, credentials
    ):

        settings = {
            "GC_PROJECT_ID": project_id,
            "GC_CREDENTIALS": credentials,
            "MIDDLEWARE": ["rele.contrib.FlaskMiddleware"],
        }

        rele.setup(settings, foo="bar")
        assert mock_middleware_setup.called
        mock_middleware_setup.call_args_list[0][-1] == {"foo": "bar"}

```

andrewgy8:
Also dont forget to add yourself to the contributors file.

PR #132, opened 2019-11-27
Clean up rele.config.setup + Worker() init

hobochild:
This relates to #114 & #119

This makes makes all config variables nullable falling back to standard
google envars, without breaking the current api.

The new apis would look like this is you have
`GOOGLE_APPLICATION_CREDENTIALS` set.

```
rele.config.setup()
```

```
w = Worker([sub1, sub2])
w.run_forever()
```

TBH, I think reading global configs from the environment is easier to
reason about than a singleton so I'd suggest that. Any non globals should
just be passed into the instances on __init__.

### :tophat: What?

Provide a description of what has been implemented.

### :thinking: Why?

Give an explanation of why.

### :link: Related issue

Add related issue's number. Example: Fix #1

### DIFF
--- a/rele/config.py
+++ b/rele/config.py
@@ -1,7 +1,7 @@
 import importlib
 import os
 
-from .client import DEFAULT_ENCODER_PATH
+from .client import DEFAULT_ENCODER_PATH, get_google_defaults, DEFAULT_ACK_DEADLINE
 from .middleware import register_middleware, default_middleware
 from .publishing import init_global_publisher
 from .subscription import Subscription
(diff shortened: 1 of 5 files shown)


hobochild:
:eyes: I assume your google creds are set in the CI -> https://travis-ci.org/mercadona/rele/jobs/617799555?utm_medium=notification&utm_source=github_status just a heads up to make sure that service account is locked down.

andrewgy8:
You are talking about the Google Cloud ones in the CI? Yeah, I scrambled the values in that file.

hobochild:
> You are talking about the Google Cloud ones in the CI? Yeah, I scrambled the values in that file.

Yea the `travis-ci-prod-2` one.

andrewgy8:
🤔 I have no idea where that value is from. We have no reference to it in our project. I wonder if thats from the machine that Travis is running on?

tests/test_config.py
```diff
@@ -51,6 +52,9 @@ def test_parses_all_keys(self, project_id, credentials, custom_encoder):
         assert config.credentials == credentials
         assert config.middleware == ["rele.contrib.DjangoDBMiddleware"]
 
+    @patch.dict(
+        os.environ, {},
```
andrewgy8: It seems like travis has a value returned for `credentials, project = get_google_defaults()`, what you might have to do is mock this function.

andrewgy8: [It is indeed a Travis envvar showing up](https://github.com/travis-ci/terraform-config/blob/master/gce-production-1/main.tf#L14)

rele/config.py
```diff
@@ -36,7 +38,7 @@ def encoder(self):
         return getattr(module, class_name)
 
 
-def setup(setting, **kwargs):
+def setup(setting={}, **kwargs):
```
tobami: A dict as default value for an argument is bad because the reference keeps being modified by subsequent calls.

hobochild: I see, so best to set it to `None` and then later set it to an empty dict if it is None?

tobami: Yep!

rele/config.py
```diff
@@ -19,13 +19,15 @@ class Config:
     """
 
     def __init__(self, setting):
-        self.gc_project_id = setting.get("GC_PROJECT_ID")
-        self.credentials = setting.get("GC_CREDENTIALS")
+        credentials, project = get_google_defaults()
```
tobami: let's only do this if `GC_PROJECT_ID` is not defined

hobochild:
Any word on this @andrewgy8 ?

andrewgy8:
Hey @hobochild ! Sorry for the delay. We will be ready to merge this asap. I see there are some conflicts. Could you fix those? I can merge it today and include it in the 0.9.0 release. Thanks!

hobochild:
Great, I'll take a look today.

hobochild:
Updated @andrewgy8

Issue #136, opened 2019-12-12
Configure Publisher Client timeout


andrewgy8:
I would like to be able to configure the timeout when publishing a message in a blocking fashion. Right now, the default and hard coded way is set to 3.0 seconds. 

I propose adding a configuration so I can declare any number of seconds for the publisher. 

Ex.

```
RELE = {
	...
	'PUBLISHER_TIMEOUT': 5.0
}
```

andrewgy8:
On second thought we could even set the timeout when publishing like:

```
rele.publish('my-topic', data={'foo': 'bar'}, blocking=True, timeout=100.0)
```

tobami:
Indeed, one option does not preclude the other!

tobami:
#137 implements a global timeout, so now we only need to implement it in `.publish()` itself

andrewgy8:
#143 will be in the next release

PR #137, opened 2019-12-12
Add publishing timeout

andrewgy8:
### :tophat: What?

Configurable timeout while waiting for a successful result from the pubsub future. https://googleapis.dev/python/pubsub/1.1.0/publisher/api/futures.html?highlight=result#google.cloud.pubsub_v1.publisher.futures.Future.result

### :thinking: Why?

We would like to control the timeout for publishing.

### :link: Related issue

Addresses #136

### DIFF
--- a/rele/client.py
+++ b/rele/client.py
@@ -97,7 +97,7 @@ class Publisher:
         else:
             self._client = pubsub_v1.PublisherClient(credentials=credentials)
 
-    def publish(self, topic, data, blocking=False, **attrs):
+    def publish(self, topic, data, blocking=False, timeout=None, **attrs):
         """Publishes message to Google PubSub topic.
 
         Usage::
@@ -114,7 +114,7 @@ class Publisher:
         Usage::
 
             publisher = Publisher()
-            future = publisher.publish('topic_name', {'foo': 'bar'}, blocking=True) # noqa
+            future = publisher.publish('topic_name', {'foo': 'bar'}, blocking=True, timeout=10.0) # noqa
 
         However, it should be noted that using `blocking=True` may incur a
         significant performance hit.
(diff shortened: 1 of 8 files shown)


rele/client.py
```diff
@@ -114,7 +114,7 @@ def publish(self, topic, data, blocking=False, **attrs):
         Usage::
 
             publisher = Publisher()
-            future = publisher.publish('topic_name', {'foo': 'bar'}, blocking=True) # noqa
+            future = publisher.publish('topic_name', {'foo': 'bar'}, blocking=True, timeout=10.0) # noqa
```
antoniobusrod: What do you think about updating code usage example with the default `PUBLISHER_TIMEOUT` 3 seconds value?

```suggestion
            future = publisher.publish('topic_name', {'foo': 'bar'}, blocking=True, timeout=3.0) # noqa
```

rele/client.py
```diff
@@ -126,6 +126,7 @@ def publish(self, topic, data, blocking=False, **attrs):
         :param topic: string topic to publish the data.
         :param data: dict with the content of the message.
         :param blocking: boolean, default False.
+        :param timeout: float, default 3.0.
```
tobami: it does not exactly default to 3. Multiple things happen here:
A) The class defaults to 3
B) Global init initializes the global publisher using `PUBLISHER_TIMEOUT` but defaults to 3
C) This method, `publish`, defaults to None, in which case it falls back to the class default (A) ), which may be 3 but it may not

So what's the right description for this docstring?

andrewgy8: In this API, default is 3. I forgot the docs for the `rele.publish` shortcut which would have a different behaviour.

andrewgy8: After a review, this seems to be a bigger issue related to the arguments being called in the `publish` api not being defined. As a result, Im going to open up a separate issue and not block this PR.

Issue #135, opened 2019-12-10
Configurable number of threads per subscription


tobami:
Right now google's library defaults to `ThreadPoolExecutor` with a max of 10 threads, which may not be what you want.

There could be a `num_threads` arguments when instantiating subscriptions that passes a scheduler with an explicit number of threads.

https://googleapis.dev/python/pubsub/latest/subscriber/index.html
https://googleapis.dev/python/pubsub/latest/subscriber/api/scheduler.html

PR #138, opened 2019-12-12
Add subscription ThreadPoolExecutor

andrewgy8:
### :tophat: What?

Add custom ThreadPoolExecutor to each subscriber. It follows the pattern shown [ here.](https://googleapis.dev/python/pubsub/latest/_modules/google/cloud/pubsub_v1/subscriber/scheduler.html#ThreadScheduler) 

And also allows customer thread count per subscription. 

Note: Ill leave this as a draft since I strongly believe that code coverage can be increased. If this is the direction we want to go, then Ill add it. 

### :thinking: Why?

We want to be able to control the number of thread per subscription. 

### :link: Related issue

Addresses #135

### DIFF
--- a/rele/client.py
+++ b/rele/client.py
@@ -57,17 +57,20 @@ class Subscriber:
                 ack_deadline_seconds=self._ack_deadline,
             )
 
-    def consume(self, subscription_name, callback):
+    def consume(self, subscription_name, callback, scheduler):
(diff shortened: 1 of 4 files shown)


andrewgy8:
Messed up creating a draft PR. Opened #139 instead

PR #139, opened 2019-12-12
Add subscription ThreadPoolExecutor

andrewgy8:
### :tophat: What?

Add custom ThreadPoolExecutor to each subscriber. It follows the pattern shown [ here.](https://googleapis.dev/python/pubsub/latest/_modules/google/cloud/pubsub_v1/subscriber/scheduler.html#ThreadScheduler) 

And also allows customer thread count per subscription. 

Note: Ill leave this as a draft since I strongly believe that code coverage can be increased. If this is the direction we want to go, then Ill add it. 

### :thinking: Why?

We want to be able to control the number of thread per subscription. 

### :link: Related issue

Addresses #135

### DIFF
--- a/rele/subscription.py
+++ b/rele/subscription.py
@@ -13,14 +13,12 @@ class Subscription:
 
     """
 
-    def __init__(self, func, topic, thread_count, prefix="", suffix="", filter_by=None):
+    def __init__(self, func, topic, prefix="", suffix="", filter_by=None):
         self._func = func
         self.topic = topic
         self._prefix = prefix
         self._suffix = suffix
         self._filter_by = filter_by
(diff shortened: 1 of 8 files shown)


rele/subscription.py
```diff
@@ -53,6 +56,17 @@ def __str__(self):
     def _filter_returns_false(self, kwargs):
         return self._filter_by and not self._filter_by(kwargs)
 
+    @property
+    def scheduler(self):
```
tobami: From an abstraction point of view I think a `Subscription` doesn't have a scheduler. IMO it would be better to let `Subscriber.consume` deal with the initialization of the scheduler based on the desired number of threads you want to use to consume

PR #142, opened 2019-12-21
Dont crash when subscription topic does not exist

### LINKED ISSUE #133: Worker should not crash when a topic does not exist
When a worker starts it tries to create subscriptions for all topics. If a topic does not exist the worker will crash.

Proposed change:
catch the `Rendezvous` exception (404 Resource not found), log an error but gracefully create and consume the rest of the defined subscriptions

andrewgy8:
### :tophat: What?

Catch google.api_core.exceptions.NotFound exception

### :thinking: Why?

Worker should not crash if the topic does not exist. 

### :link: Related issue

Fix #133

### DIFF
--- a/rele/client.py
+++ b/rele/client.py
@@ -74,7 +74,7 @@ class Subscriber:
                     ack_deadline_seconds=self._ack_deadline,
                 )
             except exceptions.NotFound:
-                logger.exception("Cannot subscribe to a topic that does not exist.")
+                logger.error("Cannot subscribe to a topic that does not exist.")
 
     def consume(self, subscription_name, callback, scheduler):
         """Begin listening to topic from the SubscriberClient.
--- a/rele/client.py
+++ b/rele/client.py
@@ -67,11 +67,14 @@ class Subscriber:
         topic_path = self._client.topic_path(self._gc_project_id, topic)
 
         with suppress(exceptions.AlreadyExists):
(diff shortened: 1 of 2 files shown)


rele/client.py
```diff
@@ -67,11 +67,14 @@ def create_subscription(self, subscription, topic):
...
-                ack_deadline_seconds=self._ack_deadline,
-            )
+            try:
+                self._client.create_subscription(
+                    name=subscription_path,
+                    topic=topic_path,
+                    ack_deadline_seconds=self._ack_deadline,
+                )
+            except exceptions.NotFound:
+                logger.exception("Cannot subscribe to a topic that does not exist.")
```
tobami: Since we are handling a very specific exception and case, we can use ERROR logging level instead of exception here. I'd also add the topic name as an interpolated variable to the log message

tests/test_client.py
```diff
@@ -92,3 +93,34 @@ def test_creates_subscription_with_custom_ack_deadline_when_provided(
...
+        )
+
+        _mocked_client.assert_called()
+
+    @patch.object(
+        SubscriberClient,
+        "create_subscription",
+        side_effect=exceptions.NotFound("Subscription topic does not exist"),
+    )
+    def test_logs_when_subscription_topic_does_not_exist(
```
tobami: `test_logs_error_`?

PR #143, opened 2019-12-21
Configure timeout from .publish()

andrewgy8:
### :tophat: What?

Configure timeout from .publish()

### :thinking: Why?

More granular timeout configuration

### :link: Related issue

Fixes #137

### DIFF
--- a/rele/client.py
+++ b/rele/client.py
@@ -158,6 +158,6 @@ class Publisher:
         if not blocking:
             return future
 
-        future.result(timeout=timeout if timeout else self._timeout)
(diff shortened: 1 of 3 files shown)


rele/client.py
```diff
@@ -157,6 +158,6 @@ def publish(self, topic, data, blocking=False, **attrs):
         if not blocking:
             return future
 
-        future.result(timeout=self._timeout)
+        future.result(timeout=timeout if timeout else self._timeout)
```
tobami:
```suggestion
        future.result(timeout=timeout or self._timeout)
```

Issue #134, opened 2019-12-10
Define multiple filters in a subscription


antoniobusrod:
`Subscription` class accepts `filter_by` to be a function that filters the messages to be processed by the sub regarding their attributes.

However, in more than occasion, if a message contains more than one attribute, it would be nice to use a list of filters for each attribute, for example.

I would like to suggest that `filter_by` parameter accepts a list of functions as well.

andrewgy8:
As of now, couldnt you just expand the one filter? Originally you might have
```
def filter_by_foo(kwargs):
    return kwargs.get('foo') == 'baz'
```
But then you want to add another 
```
def filter_by_foo_and_bar(kwargs):
    return (kwargs.get('foo') == 'baz' and kwargs.get('bar') == 'boo')
```

Its not very elegant, but in the meantime that should work. 

I agree that more elegant filtering would be something like

```
def filter_by_foo(kwargs):
    return kwargs.get('foo') == 'awesome'

def filter_by_bar(kwargs):
    return kwargs.get('bar') == 'boo'

@sub('some-topic-name', filter_by=(filter_by_foo, filter_by_bar))
def your_subscription(payload):
   ...
```

PR #148, opened 2020-01-08
 Allow multiple filters

jonasae:
### :tophat: What?

Modify subscription constructor to allow iterables in `filter_by` and apply all filters.
 
### :thinking: Why?

Current version only allows one filter.

### :link: Related issue

Solver #134

### DIFF
--- a/rele/subscription.py
+++ b/rele/subscription.py
@@ -43,7 +43,7 @@ class Subscription:
     def filter_by(self):
         return self._filters
 
-    def set_filter_by(self, filter_by):
+    def set_filters(self, filter_by):
         self._filters = filter_by
 
     def __call__(self, data, **kwargs):
--- a/rele/subscription.py
+++ b/rele/subscription.py
@@ -62,11 +62,7 @@ class Subscription:
         if not self._filters:
             return False
 
-        for filter in self._filters:
-            if not filter(kwargs):
-                return True
-
-        return False
+        return not all(filter(kwargs) for filter in self._filters)
 
 
 class Callback:
--- a/rele/subscription.py
+++ b/rele/subscription.py
@@ -154,9 +154,9 @@ def sub(topic, prefix=None, suffix=None, filter_by=None):
     :param suffix: string An optional suffix to the subscription name.
                    Useful when you have two subscribers in the same project
                    that are subscribed to the same topic.
-    :param filter_by: function An optional function that
-                      filters the messages to be processed
-                      by the sub regarding their attributes.
+    :param filter_by: Union[function, list] An optional function or tuple of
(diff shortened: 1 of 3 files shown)


jonasae:
> What do you think about documentation? I already started with the issue, maybe you find worthy next chunk.
> 
> ```diff
> diff --git a/rele/subscription.py b/rele/subscription.py
> index 73f6b03..3c4c276 100644
> --- a/rele/subscription.py
> +++ b/rele/subscription.py
> @@ -139,9 +139,9 @@ def sub(topic, prefix=None, suffix=None, filter_by=None):
>      :param suffix: string An optional suffix to the subscription name.
>                     Useful when you have two subscribers in the same project
>                     that are subscribed to the same topic.
> -    :param filter_by: function An optional function that
> -                      filters the messages to be processed
> -                      by the sub regarding their attributes.
> +    :param filter_by: Union[function, list] An optional function or tuple of
> +                      functions that filters the messages to be processed by
> +                      the sub regarding their attributes.
>      :return: :class:`~rele.subscription.Subscription`
>      """
>  
>  
> ```

Sure. I realised the documentation should be updated when I was about to finish.
Should I update de docs first in a separated PR or do it in this one?

rele/subscription.py
```diff
@@ -33,25 +41,32 @@ def set_prefix(self, prefix):
...
 
         return self._func(data, **kwargs)
 
     def __str__(self):
         return f"{self.name} - {self._func.__name__}"
 
-    def _filter_returns_false(self, kwargs):
-        return self._filter_by and not self._filter_by(kwargs)
+    def _any_filter_returns_false(self, kwargs):
+        if not self._filters:
```
tobami: Would this work instead?
```python
return not all(filter(kwargs) for filter in self._filters)
```

rele/subscription.py
```diff
@@ -33,25 +41,32 @@ def set_prefix(self, prefix):
 
     @property
     def filter_by(self):
-        return self._filter_by
+        return self._filters
 
     def set_filter_by(self, filter_by):
```
tobami: Seems to me this method should now be called `set_filters`

jonasae: Same here, but kept it because parameter is still called `filter_by`.

Changing it.

rele/subscription.py
jonasae: Almost. Too good to be `True`.

Replacing the `for` loop.

tobami: > Almost. Too good to be `True`.

:joy: 

An option to make it possible is for `_init_filters` to set `self._filters`  to empty list by default :smile:

PR #150, opened 2020-01-16
Add unrecoverable_middleware

hobochild:
This is useful if you know you will never be able to handle the message
and don't have a dead-letter queue setup.

### :tophat: What?

A custom exception and middleware to automate and ignore bad messages.

### :thinking: Why?

I've been using this a fair amount in some of our services and thought it may be useful to others.

### DIFF
--- a/rele/contrib/unrecoverable_middleware.py
+++ b/rele/contrib/unrecoverable_middleware.py
@@ -8,6 +8,4 @@ class UnrecoverableException(Exception):
 class UnrecoverableMiddleWare(BaseMiddleware):
     def post_process_message_failure(self, subscription, err, start_time, message):
         if isinstance(err, UnrecoverableException):
-            # TODO could add a config or seperate middleware
-            # to push this to a new deadletter queue
             message.ack()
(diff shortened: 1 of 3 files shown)


rele/contrib/unrecoverable_middleware.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class UnrecoverableException(Exception):
+    pass
+
+
+class UnrecoverableMiddleWare(BaseMiddleware):
+    def post_process_message_failure(self, subscription, err, start_time, message):
+        if isinstance(err, UnrecoverableException):
+            # TODO could add a config or seperate middleware
```
tobami: I do think pushing this to a deadletter queue should be done separately.

hobochild:
> This looks like a useful feature. It could even be integrated in the core subscription handling instead of as a middleware, what do you think?

tobami:
> Good point, I don't mind much either way, but I think middleware is a nice abstraction that makes it easy to see track different behaviour without diving into core logic.

OK, let's keep it as a middleware. Would you mind removing the inline comment about dead letter and adding an entry in the documentation for usage of the new exception?

hobochild:
Updated @tobami

PR #152, opened 2020-02-01
Adjust default THREADS_PER_SUBSCRIPTION

andrewgy8:
### :tophat: What?

Change [THREADS_PER_SUBSCRIPTION](https://mercadonarele.readthedocs.io/en/latest/api/settings.html#threads-per-subscription) to 2.

### :thinking: Why?

Default thread count is quite high for each subscriber at 10. It would be sensible to opt-in to a higher thread count if necessary.

### DIFF
--- a/rele/config.py
+++ b/rele/config.py
@@ -35,7 +35,7 @@ class Config:
         )
         self._encoder_path = setting.get("ENCODER_PATH", DEFAULT_ENCODER_PATH)
         self.publisher_timeout = setting.get("PUBLISHER_TIMEOUT", 3.0)
-        self.threads_per_subscription = setting.get("THREADS_PER_SUBSCRIPTION", 3)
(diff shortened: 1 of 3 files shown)


rele/config.py
```diff
@@ -35,7 +35,7 @@ def __init__(self, setting):
         )
         self._encoder_path = setting.get("ENCODER_PATH", DEFAULT_ENCODER_PATH)
         self.publisher_timeout = setting.get("PUBLISHER_TIMEOUT", 3.0)
-        self.threads_per_subscription = setting.get("THREADS_PER_SUBSCRIPTION", 10)
+        self.threads_per_subscription = setting.get("THREADS_PER_SUBSCRIPTION", 3)
```
tobami: 3? Are you a rebel against the age-old computer science preference for multiples of 2? :smile:

andrewgy8: Haha No Im not a rebel in this regard. If we prefer 2, I can change it to 2.

PR #156, opened 2020-02-21
🚨 Add Black line length settings and apply them

daaain:
### :tophat: What?

Currently Black isn't enforcing line length set at 79 as it doesn't care about `.editorconfig`. This PR adds the missing `pyproject.toml` file with Black's settings.

### :thinking: Why?

Line lengths are not corrected by Black and it also doesn't flag them up when linting.

---

I personally would prefer longer lines, so happy to rework this by changing things the other way round and updating the setting in `.editorconfig` to match Black's default (88 as far as I remember).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,4 +1,4 @@
 [tool.black]
-line-length = 79
+line-length = 88
 target_version = ['py38']
 include = '\.pyi?$'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,4 +1,4 @@
 [tool.black]
 line-length = 88
-target_version = ['py38']
+target_version = ['py36']
 include = '\.pyi?$'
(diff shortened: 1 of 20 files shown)


pyproject.toml
```diff
@@ -0,0 +1,4 @@
+[tool.black]
+line-length = 79
```
andrewgy8:
```suggestion
line-length = 88
```

How about 88?

daaain: Sounds good, I updated the line length settings everywhere, so it's consistently 88 now. So after re-running Black on all files, this means there are far fewer updates, the only ones are pretty much just the trailing commas which Black adds now that it's set to support Python 3.6+ only (which happens to be the same as Relé as far as I understood).

andrewgy8:
Thanks @daaain !

PR #157, opened 2020-02-21
🐳 Add Dockerfile and commands in Makefile

daaain:
### :tophat: What?

I set up a Dockerfile and commands in Makefile to make it possible to run all tests with a single command and no installation required (if you already have Docker).

### :thinking: Why?

I really detest installing tons of development environments on my laptop, so set this up for myself, but thought why not share.

But in general, spinning things up in a Docker container greatly lowers the barrier to entry for casual contribution.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM python:3.8-buster
+
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+WORKDIR /rele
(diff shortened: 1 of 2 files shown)


daaain:
> Perhaps we can add some documentation to the readme in regards to getting the project working locally for potential contributors. What do you think?

Makes sense, will do 👍

PR #160, opened 2020-05-25
CLI

### LINKED ISSUE #153: SUB_PREFIX is not added to subscriptions
When running Relé as a standalone worker (ie, no django, flask, etc.) I expect the subscribers to be prefixed with the SUB_PREFIX value. 

However, this does not happen since we are manually registering the subs in the worker. 

This can result in a conflicts when having multiple workers subscribed to the same topic, leading to race conditions when consuming a message.

andrewgy8:
### :tophat: What?

Provide a CLI for Relé

### :thinking: Why?

- Easily run a worker.
- Auto Discovery of subs. 
- Auto Discovery of settings.
- Ability to add more functionality to the CLI (ie. graph...)

### :link: Related issue

Fixes #153 

Here is an example on my local test application. 

Screen Shot 2020-05-25 at 9 25 29 AM
Screen Shot 2020-05-25 at 9 24 56 AM

Note: This doesnt include tests or documentation. If this feature is something we want, I will continue adding the necessary things to get it merged including tests, and documentation.


rele/__main__.py
```diff
@@ -0,0 +1,37 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def main():
+    # modify path so we can import modules and packages
+    sys.path.insert(0, os.getcwd())
+
+    parser = argparse.ArgumentParser(
+        prog="Relé", description="Unharness the power of Relé from the command line"
```
tobami, line 18: "UNharness"?

andrewgy8: Unleash the beast!

tobami: to harness its power! ;-)

rele/__main__.py
```diff
@@ -0,0 +1,39 @@
...
+        "current path. Auto-discovery will include all subs "
+        "and settings modules. If no settings module is discovered, "
+        "defaults will be used.",
+    )
+
+    args = parser.parse_args()
+
+    if args.command == "run":
+        settings, module_paths = discover.sub_modules()
+        configuration = config.setup(settings.RELE if settings else None)
```
jonasae, line 35: Why not checking if `settings.RELE` exists?

andrewgy8: Thanks @jonasae . The reason why, is that `settings` is no longer necessary :magic: If no settings are provided, we fall back to defaults. This opens up the possibility of having settings and configurations for a project in yaml, cfg. whatever format we can parse.

PR #164, opened 2020-06-08
Add isort

luisgc93:
### :tophat: What?
Add `isort` to the `requirements/test.txt` and `make lint` command. Also fix lint with isort.
### :thinking: Why?
We want to catch unsorted imports with the lint command

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ clean-tests: ## remove pytest artifacts
 
 lint: ## check style with flake8
 	black . --check --diff
+	isort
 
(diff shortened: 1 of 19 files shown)


Makefile
```diff
@@ -23,6 +23,7 @@ clean-tests: ## remove pytest artifacts
 
 lint: ## check style with flake8
 	black . --check --diff
+	isort
```
andrewgy8, line 26:
```suggestion 
	isort --check-only
```

PR #166, opened 2020-06-11
Add settings path to the cli

andrewgy8:
### :tophat: What?

Add the `--settings` path argument to the cli

### :thinking: Why?

Ive noticed different ways of configuring a project. Some with the settings in the root of the project and some projects with a settings file one level deep. This way, we can specify it anywhere and also avoid the possibility of 2 settings files in a project. 

### :link: Related issue

#160 

```
$ rele-cli run --help                  
usage: Relé run [-h] [--settings SETTINGS]

optional arguments:
  -h, --help            show this help message and exit
  --settings SETTINGS, -s SETTINGS
                        Settings file dot path. Ex. project.settings. If none
                        is supplied, Relé will attempt to autodiscover in the
                        root path.

```

```
$ rele-cli run --settings test.settings
 * Discovered subs module: 'test.subs'
 * Discovered settings: 'test.settings'
Configuring worker with 1 subscription(s)...
  test-foo-created - foo_sub
```

### DIFF
--- a/rele/__main__.py
+++ b/rele/__main__.py
@@ -27,12 +27,12 @@ def main():
         "defaults will be used.",
     )
     run_parser.add_argument(
-        '--settings',
-        '-s',
+        "--settings",
+        "-s",
         default=None,
         required=False,
-        help='Settings file dot path. Ex. project.settings. '
-             'If none is supplied, Relé will attempt to autodiscover in the root path.'
+        help="Settings file dot path. Ex. project.settings. "
+        "If none is supplied, Relé will attempt to autodiscover in the root path.",
     )
     args = parser.parse_args()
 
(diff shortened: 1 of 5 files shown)


rele/__main__.py
```diff
@@ -26,11 +26,18 @@ def main():
         "and settings modules. If no settings module is discovered, "
         "defaults will be used.",
     )
-
+    run_parser.add_argument(
+        "--settings",
+        "-s",
+        default=None,
```
tobami, line 32: Isn't the default `settings` in the current path?

andrewgy8: Surprisingly no. 😄 We do not actually need a settings path anymore. Since #132 was introduced, all other settings are either default or in the case of credentials, taken from the Google environment variables.

Issue #161, opened 2020-05-29
Autodiscovery support for class based Subscriptions


andrewgy8:
Ive seen the use of class based subscriptions become popular. 

And unfortunately, the way we autodiscover subscriptions is using [`isinstance(attribute, Subscription)`](https://github.com/mercadona/rele/blob/master/rele/config.py#L66) which does not pick up the class based subscriptions since they are not instantiated at import time, unlike when using `@sub`.

One solution that I have been thinking about is using `isinstance(attribute, Subscription) or issubclass(attribute, Subscription)`.

andrewgy8:
This is now fixed in #168

PR #168, opened 2020-06-15
Load class based subscribers

andrewgy8:
### :tophat: What?

Support class based subscribers

### :thinking: Why?

We have some services using class based subscriptions, as it removes a lot of boilerplate. 

### :link: Related issue

#161

### DIFF
--- a/rele/__init__.py
+++ b/rele/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.12.0"
+__version__ = "0.13.0"
 default_app_config = "rele.apps.ReleConfig"
 
 from .client import Publisher, Subscriber  # noqa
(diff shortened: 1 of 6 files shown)


rele/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "0.12.0"
+__version__ = "0.13.0"
```
tobami, line 1: `0.13.0dev`?

tests/subs.py
```diff
@@ -1,6 +1,25 @@
-from rele import sub
+from rele import Subscription, sub
 
 
 @sub(topic="some-cool-topic", prefix="rele")
 def sub_stub(data, **kwargs):
     return data["id"]
+
+
+class KlassBasedSub(Subscription):
```
tobami, line 9: no need to use "K" when name is more that just "class"?

Issue #159, opened 2020-05-12
Catch exceptions while waiting for futures result


andrewgy8:
Our current way of publishing while `blocking`, or waiting, for a future could be a bit dangerous. This is because Google can raise errors like a [`TimeoutError`](https://github.com/python/cpython/blob/3.8/Lib/concurrent/futures/_base.py#L441). 

I would suggest to catch the error, and add another hook called `post_publish_failure`. 

In addition, we could add another parameter to publishing which would be like `raise_exception: bool`. If `True`, we wait but if the exception is raised, we silently fail and return a bool. If `False`, we can raise the exception.

luisgc93:
> In addition, we could add another parameter to publishing which would be like raise_exception: bool. If True, we wait but if the exception is raised, we silently fail and return a bool. If False, we can raise the exception.

Wouldn't the opposite be more intuitive? So if `raise_exception` is `True` we should raise the exception and if `False` we fail silently 🤔

PR #172, opened 2020-06-28
Catch TimeoutError and run post_publish_failure when blocking

luisgc93:
### :tophat: What?

Catch `TimeoutError` when publishing while `blocking` and run `post_publish_failure` hook. Also added `raise_exception` parameter.
### :thinking: Why?
Google can raise this error when publishing so it should be handled
### :link: Related issue
#159

### DIFF
--- a/rele/client.py
+++ b/rele/client.py
@@ -171,8 +171,7 @@ class Publisher:
             run_middleware_hook("post_publish_failure", topic, e, data)
             if raise_exception:
                 raise e
-            else:
-                return False
+            return False
 
         run_middleware_hook("post_publish", topic)
         return future
--- a/rele/client.py
+++ b/rele/client.py
@@ -171,7 +171,7 @@ class Publisher:
             run_middleware_hook("post_publish_failure", topic, e, data)
             if raise_exception:
                 raise e
-            return False
+            return future
 
         run_middleware_hook("post_publish", topic)
         return future
--- a/rele/client.py
+++ b/rele/client.py
@@ -171,7 +171,7 @@ class Publisher:
             run_middleware_hook("post_publish_failure", topic, e, data)
             if raise_exception:
                 raise e
-            return future
+        else:
+            run_middleware_hook("post_publish", topic)
 
-        run_middleware_hook("post_publish", topic)
         return future
--- a/rele/client.py
+++ b/rele/client.py
@@ -166,6 +166,7 @@ class Publisher:
             future.result(timeout=timeout or self._timeout)
         except TimeoutError as e:
             run_middleware_hook("post_publish_failure", topic, e, data)
+            raise e
 
         run_middleware_hook("post_publish", topic)
         return future
--- a/rele/client.py
+++ b/rele/client.py
@@ -120,7 +120,8 @@ class Publisher:
         else:
             self._client = pubsub_v1.PublisherClient(credentials=credentials)
 
-    def publish(self, topic, data, blocking=False, timeout=None, **attrs):
+    def publish(self, topic, data, blocking=False, timeout=None, raise_exception=True,
+                **attrs):
         """Publishes message to Google PubSub topic.
 
         Usage::
@@ -146,6 +147,7 @@ class Publisher:
         message attrs using `epoch floating point number
         <https://docs.python.org/3/library/time.html#time.time>`_.
 
+        :param raise_exception: boolean. If True, exceptions coming from PubSub will be raised
         :param topic: string topic to publish the data.
         :param data: dict with the content of the message.
         :param blocking: boolean
@@ -166,7 +168,10 @@ class Publisher:
             future.result(timeout=timeout or self._timeout)
         except TimeoutError as e:
             run_middleware_hook("post_publish_failure", topic, e, data)
-            raise e
+            if raise_exception:
+                raise e
+            else:
+                return False
 
         run_middleware_hook("post_publish", topic)
         return future
--- a/rele/client.py
+++ b/rele/client.py
@@ -2,12 +2,12 @@ import json
 import logging
 import os
 import time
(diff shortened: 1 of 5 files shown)


rele/client.py
```diff
@@ -161,6 +161,10 @@ def publish(self, topic, data, blocking=False, timeout=None, **attrs):
         if not blocking:
             return future
 
-        future.result(timeout=timeout or self._timeout)
+        try:
+            future.result(timeout=timeout or self._timeout)
+        except TimeoutError as e:
```
andrewgy8, line 166: Does it raise a builtin Timeout or its own custom Timeout that we need to catch? IIRC, it is the latter case.

rele/middleware.py
```diff
@@ -47,6 +47,11 @@ def post_publish(self, topic):
         :param topic:
         """
 
+    def post_publish_failure(self, topic):
+        """Called after Publisher fails.
+        :param topic:
```
andrewgy8, line 52: Interesting. I wonder if its worth to pass in the Timeout exception to the hook, so that way we can either silence the exception, or raise with the original stack trace. What do you think?

rele/client.py
luisgc93: You're right. It comes from concurrent.futures:
https://github.com/googleapis/python-pubsub/blob/e204b86e34346b6d456771ef4d4cfe5e15e53238/google/cloud/pubsub_v1/exceptions.py#L17

rele/middleware.py
luisgc93: Actually I think we should yeah. I've noticed we also pass the exception in `post_process_message_failure`.

rele/client.py
```diff
@@ -120,7 +120,8 @@ def __init__(self, gc_project_id, credentials, encoder, timeout):
         else:
             self._client = pubsub_v1.PublisherClient(credentials=credentials)
 
-    def publish(self, topic, data, blocking=False, timeout=None, **attrs):
+    def publish(self, topic, data, blocking=False, timeout=None, raise_exception=True,
+                **attrs):
```
andrewgy8: This is strange indenting. Is this on purpose?

rele/client.py
```diff
@@ -166,6 +168,10 @@ def publish(self, topic, data, blocking=False, timeout=None, **attrs):
             future.result(timeout=timeout or self._timeout)
         except TimeoutError as e:
             run_middleware_hook("post_publish_failure", topic, e, data)
```
andrewgy8, line 171: We should add logger.exception to the LoggingMiddleware on this hook.

tests/test_client.py
```diff
@@ -77,7 +76,53 @@ def test_publishes_data_with_client_timeout_when_blocking_and_timeout_specified(
...
             ANY, b'{"foo": "bar"}', published_at=ANY
         )
-        mock_future.assert_called_once_with(timeout=50)
+        mock_future.result.assert_called_once_with(timeout=50)
+
+    def test_runs_post_publish_failure_hook_when_time_out_error(
+        self, mock_future, publisher, mock_post_publish_failure
+    ):
+        message = {"foo": "bar"}
+        e = TimeoutError()
```
andrewgy8, line 85:
```suggestion
        exception = TimeoutError()
```

tests/test_client.py
```diff
@@ -77,7 +76,53 @@ def test_publishes_data_with_client_timeout_when_blocking_and_timeout_specified(
         publisher._client.publish.assert_called_with(
             ANY, b'{"foo": "bar"}', published_at=ANY
         )
-        mock_future.assert_called_once_with(timeout=50)
+        mock_future.result.assert_called_once_with(timeout=50)
+
+    def test_runs_post_publish_failure_hook_when_time_out_error(
```
andrewgy8, line 81:
```suggestion
    def test_runs_post_publish_failure_hook_when_future_result_raises_timeout(
```

tests/test_client.py
```diff
@@ -77,7 +76,53 @@ def test_publishes_data_with_client_timeout_when_blocking_and_timeout_specified(
...
+        mock_future.result.side_effect = e
+
+        result = publisher.publish(
+            topic="order-cancelled",
+            data=message,
+            myattr="hello",
+            blocking=True,
+            raise_exception=False,
+        )
+        assert result is False
```
andrewgy8, line 125: What do you think about returning the `future` here instead?

rele/client.py
```diff
@@ -161,6 +165,14 @@ def publish(self, topic, data, blocking=False, timeout=None, **attrs):
...
 
-        future.result(timeout=timeout or self._timeout)
+        try:
+            future.result(timeout=timeout or self._timeout)
+        except TimeoutError as e:
+            run_middleware_hook("post_publish_failure", topic, e, data)
+            if raise_exception:
+                raise e
+            else:
+                return False
```
jonasae, line 175: What do you think about this to get rid of the `else`?
```suggestion
            if raise_exception:
                raise e
                
            return False
```

rele/client.py
luisgc93: I fixed it in [this commit](https://github.com/mercadona/rele/pull/172/commits/f8ad956e42b608f066887be23dd7eed748859231) but for some reason the changes are not showing up in the files changed section. I've also added locally a pre-commit hook with a lint check to make sure this doesn't happen again.

rele/contrib/logging_middleware.py
```diff
@@ -48,6 +48,20 @@ def pre_publish(self, topic, data, attrs):
             },
         )
 
+    def post_publish_failure(self, topic, exception, message):
+        self._logger.exception(
+            f"Exception raised while processing message "
```
antoniobusrod, line 53: What do you think to relate the log content to the "publish" failure event?

```suggestion
            f"Exception raised while publishing message "
```

Issue #176, opened 2020-07-15
Remove PROJECT_ID


andrewgy8:
Since we are using the credentials object in the project, we can extract the Project id from the credentials. This would eliminate the need for another settings configuration.

PR #178, opened 2020-07-17
Deprecate GC_PROJECT_ID setting

andrewgy8:
### :tophat: What?

Deprecate GC_PROJECT_ID. I was even thinking about removing it. Since we can fallback to using the credentials, the removal shouldnt be a breaking change. What do you all think? 

### :thinking: Why?

Simpler setup or a user. Less settings == less hassle

### :link: Related issue

#176

### DIFF
--- a/rele/client.py
+++ b/rele/client.py
@@ -32,6 +32,7 @@ class Subscriber:
     For convenience, this class wraps the creation and consumption of a topic
     subscription.
 
+    :param gc_project_id: string :ref:`_settings_project_id` .
     :param credentials: obj :meth:`~rele.config.Config.credentials`.
     :param default_ack_deadline: int Ack Deadline defined in settings
     """
--- a/rele/client.py
+++ b/rele/client.py
@@ -32,19 +32,18 @@ class Subscriber:
     For convenience, this class wraps the creation and consumption of a topic
     subscription.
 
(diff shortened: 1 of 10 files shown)


rele/client.py
```diff
@@ -31,24 +31,19 @@ class Subscriber:
...
     :param default_ack_deadline: int Ack Deadline defined in settings
     """
 
-    def __init__(self, gc_project_id=None, credentials=None, default_ack_deadline=None):
-
-        if gc_project_id is None or credentials is None:
-            creds, project = get_google_defaults()
-
-        self._gc_project_id = gc_project_id or project
+    def __init__(self, gc_project_id, credentials, default_ack_deadline=None):
```
tobami, line 39: if we are deprecating it, this parameter should stay optional

andrewgy8: We are deprecating it from the settings. However from the project, like the subscriber client, I am afraid we can not remove it. 
I did decide to remove the non-default values here since we are no longer taking the credentials from the environment in the client. Instead we take the values from the environment in the config.

andrewgy8: Basically, instead of having credentials being instantiated in multiple places, I have consolidated the handling of credentials in the Config class, no where else.

PR #187, opened 2020-10-01
Fix #180: Raise error when the config loads a repeated subscription

### LINKED ISSUE #180: Worker initialization should raise an error on duplicate subscriptions
If two `subs` subscribe to the same topic and don't declare a suffix, stealing will happen where approximately half the messages will be processed by one subscription and half by the other. That's extremely difficult to test against and to debug in production, so a protection would be very useful.

I'd suggest raising an error straight away so the worker doesn't start. A warning log wouldn't be as effective because it could go unnoticed.

EmilioCarrion:
### :tophat: What?

Raise error when the config loads a repeated subscription

### :thinking: Why?

The issue describes it well:
https://github.com/mercadona/rele/issues/180

Fix #180

### DIFF
--- a/rele/config.py
+++ b/rele/config.py
@@ -103,9 +103,11 @@ def load_subscriptions_from_paths(sub_module_paths, sub_prefix=None, filter_by=N
                 subscription.set_filters(filter_by)
 
             if subscription.name in subscriptions:
+                found_subscription = subscriptions[subscription.name]
                 raise RuntimeError(
                     f"Duplicated subscription found: {subscription.name}. Handled by "
-                    f"{sub_module_path}.{attr_name} and some more."
+                    f"{subscription._func.__module__}.{subscription._func.__name__} and "
+                    f"{found_subscription._func.__module__}.{found_subscription._func.__name__}."
                 )
 
             subscriptions[subscription.name] = subscription
--- a/rele/config.py
+++ b/rele/config.py
@@ -87,7 +87,7 @@ def subscription_from_attribute(attribute):
 
 def load_subscriptions_from_paths(sub_module_paths, sub_prefix=None, filter_by=None):
 
-    subscriptions = []
+    subscriptions = {}
     for sub_module_path in sub_module_paths:
         sub_module = importlib.import_module(sub_module_path)
         for attr_name in dir(sub_module):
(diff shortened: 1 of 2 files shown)


EmilioCarrion:
There were two options to implement this logic. In the worker inicialization or in the load of subscriptions from the modules. I decided to implement it in the load due to that we might want to instanciate multiple workers with the subscriptions spread between them and may be repeated but not in a single Worker.

rele/config.py
```diff
@@ -102,5 +102,10 @@ def load_subscriptions_from_paths(sub_module_paths, sub_prefix=None, filter_by=N
...
                 subscription.set_filters(filter_by)
 
-            subscriptions.append(subscription)
-    return subscriptions
+            if subscription.name in subscriptions:
+                raise RuntimeError(
+                    f"Duplicated subscription found: {subscription.name}."
+                )
+
+            subscriptions[subscription.name] = subscription
```
andrewgy8, line 110: Could this be kept as a list and instead test for `subscription.name in subscriptions`?

rele/config.py
```diff
@@ -102,5 +102,10 @@ def load_subscriptions_from_paths(sub_module_paths, sub_prefix=None, filter_by=N
             if filter_by and not subscription.filter_by:
                 subscription.set_filters(filter_by)
 
-            subscriptions.append(subscription)
-    return subscriptions
+            if subscription.name in subscriptions:
+                raise RuntimeError(
```
andrewgy8, line 106: Do we want to raise or skip and log? If we skip, the behaviour will be [similar](https://github.com/mercadona/rele/blob/master/rele/client.py#L72) to what happens when there is no topic for the subscription.

tobami: I'd raise because it's not something that should be done. A duplicate sub will only bring pain(either splitting consumption or if we skip not knowing why a given sub is not executed)

rele/config.py
EmilioCarrion: Yeah, but it will iterate through the full list in every iteration to check if the item.name has already been registered. The dict aproach ensures a constant check every validation

andrewgy8 (approved):

rele/config.py
```diff
@@ -102,5 +102,10 @@ def load_subscriptions_from_paths(sub_module_paths, sub_prefix=None, filter_by=N
...
+                    f"Duplicated subscription found: {subscription.name}."
+                )
```
tobami, line 108: It would be usefule to also log the names of the two subscription `functions` that collide

EmilioCarrion: Done it in https://github.com/mercadona/rele/pull/187/commits/3b3256050aa13935c878f076df358550b00d8a89

rele/config.py
```diff
@@ -103,9 +103,11 @@ def load_subscriptions_from_paths(sub_module_paths, sub_prefix=None, filter_by=N
                 subscription.set_filters(filter_by)
 
             if subscription.name in subscriptions:
+                found_subscription = subscriptions[subscription.name]
                 raise RuntimeError(
                     f"Duplicated subscription found: {subscription.name}. Handled by "
```
tobami, line 108: I would reformulate this to:
`Duplicate subscription name found: {subscription.name]. subs {sub1} and {sub2} collide`

PR #188, opened 2020-10-01
Discover and load settings when publishing

### LINKED ISSUE #177: Publishing without boilerplate
## Current Behaviour

Right now, if you want to call `rele.publish()`, you must call `config.setup()` before. This is so that 
`init_global_publisher` is called and setup properly. Otherwise a `ValueError` is raised.

## Proposal
I propose that a user should be able to call `rele.publish` without the boilerplate of `config.setup`. Instead, I think we can call `init_global_publisher` if publishing and there is no global publisher. The tough part will be getting the settings so that the credentials can be configured properly. 

But I do believe this can be solved, to make the UX more elegant and avoid boilerplate.

EmilioCarrion:
### :tophat: What?

Implement discovering and setup of settings when publishing without calling to `config.setup()`

### :thinking: Why?

To avoid the boilerplate in simple setups.

### :link: Related issue
Fix #177

### DIFF
--- a/rele/publishing.py
+++ b/rele/publishing.py
@@ -1,3 +1,5 @@
+from rele import config, discover
+
 from .client import Publisher
 
 _publisher = None
@@ -41,5 +43,10 @@ def publish(topic, data, **kwargs):
     :return: None
     """
     if not _publisher:
-        raise ValueError("init_global_publisher must be called first.")
+        settings, _ = discover.sub_modules()
+        if not hasattr(settings, "RELE"):
+            raise ValueError("Config setup not called and settings module not found.")
+
+        config.setup(settings.RELE)
+
     _publisher.publish(topic, data, **kwargs)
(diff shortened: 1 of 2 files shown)


rele/publishing.py
```diff
@@ -41,5 +43,10 @@ def myfunc():
     :return: None
     """
     if not _publisher:
-        raise ValueError("init_global_publisher must be called first.")
+        settings, _ = discover.sub_modules()
```
andrewgy8, line 46: I know this will work for settings that are in the top level of the project dir. However, I have my doubts that it'll work for settings that are located elsewhere. Do you know if this is the case?

EmilioCarrion: Yeah, but I undestood this as a no boilerplate publish for simple setups. For a hidden settings inside the project it should be used with the config.setup. But we can change el discover to find settings inside the project.

andrewgy8 (approved):

PR #189, opened 2020-10-02
Add validations to the sub decorator

### LINKED ISSUE #186: Improve sub decorator
The `sub` decorator can get some ease of use improvements

1. Use [functools.wraps](https://docs.python.org/3/library/functools.html#functools.wraps) to preserve `__name__` and `__doc__` of the original function
2. [Inspect](https://docs.python.org/3/library/functools.html#functools.wraps) callback function signature and raise an exception if the signature is not compatible (ie. missing `data`)
3. Log a warning message if the function cannot be discovered (ie. no `sub` in the path)

Slightly related to the second point: `@sub("topic", filter_by=42)` will raise a `TypeError` at runtime (because filters are used as callables but they are not checked)

EmilioCarrion:
### :tophat: What?

Add validations to the sub decorator. 
Validate the signature of the subscription function to check if it has the correct number of arguments.
Validate the `filter_by` parameter to check if it is a callable or a iterable of callables as defined in the docstring.
Log a warning if the subscription function it's outside a discoverable module to alert the user that that susbcription will not be created.

### :thinking: Why?

Improvements suggested in https://github.com/mercadona/rele/issues/186

The use of the `wraps` decorator in the sub decorator is not useful as the decorator returns a Subscription instance and not a function so the expected behaviour of keep the `__name__` and `__doc__` is lost. If someone knows the way to acomplish it please leave a comment.

### :link: Related issue

Fix #186

### DIFF
--- a/rele/subscription.py
+++ b/rele/subscription.py
@@ -42,7 +42,16 @@ class Subscription:
         self._filters = self._init_filters(filter_by)
 
     def _init_filters(self, filter_by):
-        if hasattr(filter_by, "__iter__"):
+        if filter_by and not (
+            callable(filter_by)
+            or (
+                isinstance(filter_by, Iterable)
+                and all(callable(filter) for filter in filter_by)
+            )
+        ):
+            raise ValueError("Filter_by must be a callable or a list of callables.")
+
+        if isinstance(filter_by, Iterable):
             return filter_by
         elif filter_by:
             return [filter_by]
@@ -180,18 +189,10 @@ def sub(topic, prefix=None, suffix=None, filter_by=None):
                       the sub regarding their attributes.
     :return: :class:`~rele.subscription.Subscription`
     """
-    if filter_by and not (
-        callable(filter_by)
-        or (
-            isinstance(filter_by, Iterable)
-            and all(callable(filter) for filter in filter_by)
-        )
-    ):
-        raise ValueError("Filter_by must be a callable or a list of callables.")
 
     def decorator(func):
         args_spec = getfullargspec(func)
-        if len(args_spec.args) < 2:
+        if len(args_spec.args) != 1:
             raise RuntimeError(
                 f"Subscription function {func.__module__}.{func.__name__} is not valid. "
                 "The function must have one argument. "
--- a/rele/subscription.py
+++ b/rele/subscription.py
@@ -1,6 +1,8 @@
 import json
 import logging
 import time
+from collections.abc import Iterable
+from inspect import getfullargspec, getmodule, signature
 
 from .middleware import run_middleware_hook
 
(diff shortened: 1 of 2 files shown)


tests/test_subscription.py
```diff
@@ -308,3 +308,32 @@ def test_old_django_connections_closed_when_middleware_is_used(
...
+    def test_raises_error_when_function_signature_is_not_valid(self):
+        with pytest.raises(RuntimeError):
+            sub(topic="topic", prefix="rele")(lambda: None)
+
+    def test_logs_warning_when_function_not_in_subs_module(self, caplog):
+        sub(topic="topic", prefix="rele")(lambda data: None)
+        assert (
+            "Subscription function tests.test_subscription.<lambda> is outside a subs "
+        )
+        "module that will not be discovered." in caplog.text
```
andrewgy8, line 327: Is there something wrong with the formatting here?

tests/test_subscription.py
```diff
@@ -308,3 +308,32 @@ def test_old_django_connections_closed_when_middleware_is_used(
 
         assert res == 123
         assert mock_close_old_connections.call_count == 2
+
+
+class TestDecorator:
+    def test_returns_susbscription_when_function_valid(self):
```
andrewgy8, line 314:
```suggestion
    def test_returns_susbscription_when_callback_valid(self):
```
?

rele/subscription.py
```diff
@@ -178,8 +180,29 @@ def sub_process_landscape_photos(data, **kwargs):
...
+        or (
+            isinstance(filter_by, Iterable)
+            and all(callable(filter) for filter in filter_by)
+        )
+    ):
+        raise ValueError("Filter_by must be a callable or a list of callables.")
 
     def decorator(func):
+        args_spec = getfullargspec(func)
+        if len(args_spec.args) != 1:
```
andrewgy8, line 194: I think this should be `< 2`. We will always be passing atleast two args into the callback. One is the `data` and other(s) will be the attributes. One of them should always be present which is the `published_at` field.

rele/subscription.py
```diff
@@ -178,8 +180,29 @@ def sub_process_landscape_photos(data, **kwargs):
                       the sub regarding their attributes.
     :return: :class:`~rele.subscription.Subscription`
     """
+    if filter_by and not (
+        callable(filter_by)
+        or (
+            isinstance(filter_by, Iterable)
+            and all(callable(filter) for filter in filter_by)
+        )
+    ):
```
andrewgy8, line 189: What do think about placing this in the Subscription class. Im afraid that the sub decorator is not the only place that we can instantiate a Subscription. Now that we support class based subscriptions #168

tests/test_subscription.py
EmilioCarrion: Black lied to me! I'll change it

rele/subscription.py
EmilioCarrion: Yeah but here is checking the arguments, not the keyword arguments that are optional.

rele/subscription.py
EmilioCarrion: Moved.

andrewgy8 (approved):

PR #190, opened 2020-10-03
Add new post_publish_hook and deprecate the old one

### LINKED ISSUE #181: Rename post_publish hook to post_publish_success
We want to be more explicit with this middleware hook, just like we do with`post_process_message_failure` and `post_process_message_success`. 
For backwards compatibility we'll Initially allow both `post_publish` and `post_publish_success` and eventually we'll deprecate `post_publish`.

EmilioCarrion:
### :tophat: What?

Added new post_publish_hook and deprecate the old one

### :thinking: Why?

To have a more consistent interface as suggested in https://github.com/mercadona/rele/issues/181

### :link: Related issue

Fix #181

### DIFF
--- a/rele/middleware.py
+++ b/rele/middleware.py
@@ -62,7 +62,7 @@ class BaseMiddleware(metaclass=WarnDeprecatedHooks):
         """
 
     def post_publish_success(self, topic):
-        """DEPRECATED: Called after Publisher sends message.
+        """Called after Publisher sends message.
         :param topic:
         """
 
--- a/rele/middleware.py
+++ b/rele/middleware.py
@@ -1,4 +1,5 @@
 import importlib
+import warnings
 
 _middlewares = []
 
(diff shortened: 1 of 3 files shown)


rele/middleware.py
```diff
@@ -43,7 +57,12 @@ def pre_publish(self, topic, data, attrs):
         """
 
     def post_publish(self, topic):
-        """Called after Publisher sends message.
+        """DEPRECATED: Called after Publisher sends message.
+        :param topic:
+        """
+
+    def post_publish_success(self, topic):
+        """DEPRECATED: Called after Publisher sends message.
```
luisgc93, line 65: I imagine this `DEPRECATED` was added by mistake, right? If I remember correctly, it's meant to go under `post_publish_success` only

EmilioCarrion: Yeah, thanks!

tests/test_middleware.py
```diff
@@ -23,3 +24,11 @@ def test_setup_fn_is_called_with_kwargs(self, mock_middleware_setup, project_id)
         rele.setup(settings, foo="bar")
         assert mock_middleware_setup.called
         assert mock_middleware_setup.call_args_list[0][-1] == {"foo": "bar"}
+
+    def test_warns_about_deprecated_hooks(self, caplog):
```
andrewgy8, line 28: Do you want to assert anything with the `caplog` here?

EmilioCarrion: Uups 😊

andrewgy8 (approved):
Lets go!

PR #192, opened 2020-10-11
Google Pubsub 2.0 Compat

### LINKED ISSUE #184: Breaking with Google PubSub >= 2.0
In the 1.0 release, I had to pin pubsub to <2.0. Otherwise, our tests would break. It seems there were major changes, and we need to update some usages in our code. 

To reproduce: 
1. update google pubsub library to >2.0. 
1. Run tests

luisgc93:
### :tophat: What?

Fix breaking changes of Google PubSub >= 2.0

### :thinking: Why?
The new release introduced two breaking changes:
1. New way of retrieving the topic path (these [lines](https://github.com/googleapis/python-pubsub/commit/6b9eec81ccee81ab93646eaf7652139fc218ed36#diff-f290253b439dfbdadd108ac2e613dd1cR189-R191))
2. Added `publish_time` attribute to `Message` class (this [line](https://github.com/googleapis/python-pubsub/commit/6b9eec81ccee81ab93646eaf7652139fc218ed36#diff-9d37cc3aaac018e0c394df450a873e66R112))

### :link: Related issue

Fixes #184

### DIFF
--- a/rele/client.py
+++ b/rele/client.py
@@ -60,8 +60,8 @@ class Subscriber:
         subscription_path = self._client.subscription_path(
             self._gc_project_id, subscription
         )
-        publisher = pubsub_v1.PublisherClient()
-        topic_path = publisher.topic_path(self._gc_project_id, topic)
+        subscriber = pubsub_v1.SubscriberClient()
+        topic_path = subscriber.topic_path(self._gc_project_id, topic)
 
         with suppress(exceptions.AlreadyExists):
             try:
--- a/rele/client.py
+++ b/rele/client.py
@@ -60,8 +60,8 @@ class Subscriber:
         subscription_path = self._client.subscription_path(
             self._gc_project_id, subscription
         )
-        subscriber = pubsub_v1.SubscriberClient()
-        topic_path = subscriber.topic_path(self._gc_project_id, topic)
+        publisher = pubsub_v1.PublisherClient()
+        topic_path = publisher.topic_path(self._gc_project_id, topic)
 
         with suppress(exceptions.AlreadyExists):
             try:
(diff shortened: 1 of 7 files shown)


rele/client.py
```diff
@@ -60,7 +60,8 @@ def create_subscription(self, subscription, topic):
         subscription_path = self._client.subscription_path(
             self._gc_project_id, subscription
         )
-        topic_path = self._client.topic_path(self._gc_project_id, topic)
+        publisher = pubsub_v1.PublisherClient()
+        topic_path = publisher.topic_path(self._gc_project_id, topic)
```
andrewgy8, line 64: I wonder if its still possible to do this with the SubscriptionClient following the [docs here](https://googleapis.dev/python/pubsub/latest/subscriber/index.html#creating-a-subscription). 

If not, then we have to be very careful about memory leaks, as if IIRC, the publisher class is the one that gave us trouble.

setup.py
```diff
@@ -39,7 +39,7 @@ def get_version(*file_paths):
     url="https://github.com/mercadona/rele",
     packages=find_packages(exclude=("tests",)),
     include_package_data=True,
-    install_requires=["google-auth", "google-cloud-pubsub<2.0"],
+    install_requires=["google-auth", "google-cloud-pubsub"],
```
andrewgy8, line 42: We will have to force usage of `google-cloud-pubsub>=2.0`

andrewgy8 (changes requested):

rele/client.py
luisgc93: I'm not so sure about this now. I can't seem to find any `topic_path` method for the subscriber in the pubsub project, only the `PublisherClient` class  has it. Perhaps that section of the docs was not updated?

luisgc93: I'm not so sure about this now. I can't seem to find a `topic_path` method for the subscriber class in the pubsub project, only `PublisherClient` has it. Perhaps that section of the docs was not updated?

andrewgy8 (approved):
👏

PR #206, opened 2021-09-07
Automatic topic creation

mgarcia0094:
### :tophat: What?

Automatic topic creation when topic does not exist in PubSub.

### :thinking: Why?

When we trying to subscribe to a topic that doesn't exist, there is a crash. In this feature, we create the topic automatically if this does not exist and then retry to initialize the subscription.

PubSub credentials must have editor role.

### :link: Related issue

None

### DIFF
--- a/rele/client.py
+++ b/rele/client.py
@@ -42,9 +42,15 @@ class Subscriber:
         self._ack_deadline = default_ack_deadline or DEFAULT_ACK_DEADLINE
 
         if USE_EMULATOR:
-            self._client = pubsub_v1.SubscriberClient()
+            self._subscriber_client = pubsub_v1.SubscriberClient()
+            self._publisher_subscriber_client = pubsub_v1.PublisherClient()
         else:
-            self._client = pubsub_v1.SubscriberClient(credentials=credentials)
+            self._subscriber_client = pubsub_v1.SubscriberClient(
+                credentials=credentials
+            )
+            self._publisher_subscriber_client = pubsub_v1.PublisherClient(
+                credentials=credentials
+            )
 
     def create_subscription(self, subscription, topic):
         """Handles creating the subscription when it does not exists.
@@ -57,20 +63,27 @@ class Subscriber:
         :param subscription: str Subscription name
         :param topic: str Topic name to subscribe
         """
-        subscription_path = self._client.subscription_path(
+        subscription_path = self._subscriber_client.subscription_path(
             self._gc_project_id, subscription
         )
-        topic_path = self._client.topic_path(self._gc_project_id, topic)
+        topic_path = self._subscriber_client.topic_path(self._gc_project_id, topic)
 
         with suppress(exceptions.AlreadyExists):
             try:
-                self._client.create_subscription(
+                self._subscriber_client.create_subscription(
                     name=subscription_path,
                     topic=topic_path,
                     ack_deadline_seconds=self._ack_deadline,
                 )
             except exceptions.NotFound:
-                logger.error("Cannot subscribe to a topic that does not exist.")
+                logger.info(
+                    "Cannot subscribe to a topic that does not exist."
+                    "Creating topic..."
+                )
+                topic = self._publisher_subscriber_client.create_topic(
+                    request={"name": topic_path}
+                )
+                logger.info(f"Topic {topic.name} created.")
 
     def consume(self, subscription_name, callback, scheduler):
         """Begin listening to topic from the SubscriberClient.
@@ -80,10 +93,10 @@ class Subscriber:
         :param scheduler: `Thread pool-based scheduler. <https://googleapis.dev/python/pubsub/latest/subscriber/api/scheduler.html?highlight=threadscheduler#google.cloud.pubsub_v1.subscriber.scheduler.ThreadScheduler>`_  # noqa
         :return: `Future <https://googleapis.github.io/google-cloud-python/latest/pubsub/subscriber/api/futures.html>`_  # noqa
         """
-        subscription_path = self._client.subscription_path(
+        subscription_path = self._subscriber_client.subscription_path(
             self._gc_project_id, subscription_name
         )
-        return self._client.subscribe(
+        return self._subscriber_client.subscribe(
             subscription_path, callback=callback, scheduler=scheduler
         )
 
--- a/rele/client.py
+++ b/rele/client.py
@@ -40,15 +40,8 @@ class Subscriber:
     def __init__(self, gc_project_id, credentials, default_ack_deadline=None):
         self._gc_project_id = gc_project_id
         self._ack_deadline = default_ack_deadline or DEFAULT_ACK_DEADLINE
-
-        if USE_EMULATOR:
-            self._subscriber_client = pubsub_v1.SubscriberClient()
-            self._publisher_client = pubsub_v1.PublisherClient()
-        else:
-            self._subscriber_client = pubsub_v1.SubscriberClient(
-                credentials=credentials
-            )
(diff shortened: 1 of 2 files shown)


rele/client.py
```diff
@@ -57,20 +63,31 @@ def create_subscription(self, subscription, topic):
...
-                    topic=topic_path,
-                    ack_deadline_seconds=self._ack_deadline,
-                )
+                self._create_subscription(subscription_path, topic_path)
             except exceptions.NotFound:
-                logger.error("Cannot subscribe to a topic that does not exist.")
+                logger.info(
+                    "Cannot subscribe to a topic that does not exist."
+                    "Creating topic..."
+                )
```
daniel-ruiz, line 78: Could you please include the name of the failing topic in the logs? That is precious information when you find these messages in Sentry.

rele/client.py
```diff
@@ -42,9 +42,15 @@ def __init__(self, gc_project_id, credentials, default_ack_deadline=None):
...
+            self._subscriber_client = pubsub_v1.SubscriberClient()
+            self._publisher_subscriber_client = pubsub_v1.PublisherClient()
         else:
-            self._client = pubsub_v1.SubscriberClient(credentials=credentials)
+            self._subscriber_client = pubsub_v1.SubscriberClient(
+                credentials=credentials
+            )
+            self._publisher_subscriber_client = pubsub_v1.PublisherClient(
+                credentials=credentials
+            )
```
daniel-ruiz, line 53: The name `_publisher_subscriber_client` seems confusing to me. It looks like it's mixing opposed terms: publishers and subscribers.
Why did you choose that name? Have you thought about renaming it to `_publisher_client`? 🤔

daniel-ruiz (approved):

rele/client.py
```diff
@@ -57,20 +63,31 @@ def create_subscription(self, subscription, topic):
...
             try:
-                self._client.create_subscription(
-                    name=subscription_path,
-                    topic=topic_path,
-                    ack_deadline_seconds=self._ack_deadline,
-                )
+                self._create_subscription(subscription_path, topic_path)
             except exceptions.NotFound:
-                logger.error("Cannot subscribe to a topic that does not exist.")
+                logger.info(
```
tobami, line 75: I would log it as warning, not info. And I agree that having the topic name would be very helpful

rele/client.py
```diff
@@ -42,9 +42,15 @@ def __init__(self, gc_project_id, credentials, default_ack_deadline=None):
         self._ack_deadline = default_ack_deadline or DEFAULT_ACK_DEADLINE
 
         if USE_EMULATOR:
-            self._client = pubsub_v1.SubscriberClient()
+            self._subscriber_client = pubsub_v1.SubscriberClient()
+            self._publisher_subscriber_client = pubsub_v1.PublisherClient()
         else:
-            self._client = pubsub_v1.SubscriberClient(credentials=credentials)
+            self._subscriber_client = pubsub_v1.SubscriberClient(
```
tobami, line 48: I wouldn't touch `self._client`, it's the main private attribute of the `Subscriber` class, it doesn't deserve to become ugly just because we need to cover a corner case :grin: 

In addition to that, we should probably use the global publisher anyway, so no need to initialize a publishing client at all

tobami (changes requested):

tests/test_client.py
```diff
@@ -187,16 +194,14 @@ def test_does_not_raise_when_subscription_already_exists(
     @patch.object(
         SubscriberClient,
         "create_subscription",
-        side_effect=exceptions.NotFound("Subscription topic does not exist"),
+        side_effect=[exceptions.NotFound("Subscription topic does not exist"), True],
     )
-    def test_logs_error_when_subscription_topic_does_not_exist(
-        self, _mocked_client, project_id, subscriber, caplog
+    def test_create_topic_when_subscription_topic_does_not_exist(
```
tobami, line 199: this should be third person, as in `test_creates_topic_when_`

tests/test_client.py
```diff
@@ -187,16 +194,14 @@ def test_does_not_raise_when_subscription_already_exists(
...
         subscriber.create_subscription(
             subscription="test-topic", topic=f"{project_id}-test-topic"
         )
 
-        _mocked_client.assert_called()
-        log = caplog.records[0]
-        assert log.message == "Cannot subscribe to a topic that does not exist."
-        assert log.levelname == "ERROR"
+        assert _mocked_client.call_count == 2
+        mock_create_topic.assert_called()
```
tobami, line 207: Can we use `assert_called_with`?

rele/client.py
mgarcia0094: Done!

rele/client.py
mgarcia0094: It was a mistake. Fixed.

rele/client.py
mgarcia0094: Done.

tests/test_client.py
mgarcia0094: done!

rele/client.py
mgarcia0094: I think `self._subscriber_client` is more descriptive. Anyway, `Subscriber` class should not have dependencies with the `Publisher` class.

rele/client.py
```diff
@@ -43,14 +43,12 @@ def __init__(self, gc_project_id, credentials, default_ack_deadline=None):
...
-            self._publisher_subscriber_client = pubsub_v1.PublisherClient()
+            self._publisher_client = pubsub_v1.PublisherClient()
         else:
             self._subscriber_client = pubsub_v1.SubscriberClient(
                 credentials=credentials
             )
-            self._publisher_subscriber_client = pubsub_v1.PublisherClient(
-                credentials=credentials
-            )
+            self._publisher_client = pubsub_v1.PublisherClient(credentials=credentials)
```
tobami, line 51: Even better: let's not initialize a publisher client in the init that the subscription won't use 99.999% of the time :smile: 

We can have a `_create_topic` that initializes a local client just for that purpose

mgarcia0094: Ok! Done 😄

rele/client.py
```diff
@@ -74,12 +67,14 @@ def create_subscription(self, subscription, topic):
                     "Cannot subscribe to a topic that does not exist."
                     f"Creating {topic_path}..."
                 )
-                topic = self._publisher_client.create_topic(
-                    request={"name": topic_path}
-                )
+                topic = self._create_topic(topic_path)
                 logger.warning(f"{topic.name} created.")
```
tobami, line 71:
```suggestion
                logger.warning(f"Topic {topic.name} created.")
```

rele/client.py
```diff
@@ -40,11 +40,8 @@ class Subscriber:
     def __init__(self, gc_project_id, credentials, default_ack_deadline=None):
         self._gc_project_id = gc_project_id
         self._ack_deadline = default_ack_deadline or DEFAULT_ACK_DEADLINE
-
-        if USE_EMULATOR:
-            self._client = pubsub_v1.SubscriberClient()
-        else:
-            self._client = pubsub_v1.SubscriberClient(credentials=credentials)
+        self.credentials = credentials if not USE_EMULATOR else None
+        self._subscriber_client = pubsub_v1.SubscriberClient(credentials=credentials)
```
tobami, line 44: I still see no reason to rename `Subscriber._client` to `Subscriber._subscriber_client`. `client` profits from the class namespace. Plus now this class will have inconsistent naming compared to `Publisher`

mgarcia0094: I disagree but I have already changed it. 👍

tobami (approved):

antoniobusrod (approved):

PR #207, opened 2021-09-08
Add filter expressions to subscriptions

EmilioCarrion:
### :tophat: What?

Add filter expressions to subscriptions. 

Filter expressions are applied by pubsub before sending the message to a specific subscriber. It's use is defined [here](https://cloud.google.com/pubsub/docs/filtering#filtering_syntax) and it prevents the messages to be sent and filtered by the workers.

### :thinking: Why?

To give support to this pubsub feature.

### :link: Related issue

N/A

### DIFF
--- a/docs/guides/filters.rst
+++ b/docs/guides/filters.rst
@@ -2,8 +2,9 @@ Filtering Messages
 ==================
 
 Filter can be used to execute a subscription with specific parameters.
-There's two types of filters, global or passing a filter_by parameter in the
-subscription.
+There's three types of filters, global, by passing a filter_by parameter in the
+subscription (this applies the filter locally) or by passing a filter_expression
+parameter in the subscription (this applies the filter on pubsub).
 
 
 `filter_by` parameter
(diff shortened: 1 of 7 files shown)


docs/guides/filters.rst
```diff
@@ -26,6 +27,26 @@ is passed as parameter ``filter_by`` in the subscription.
         print(f'Received a photo of type {kwargs.get("type")}')
 
 
+`filter_expression` parameter
```
tobami, line 30: I wonder whether we should be calling this `filter_expression`. We are exploring asyncapi compatible naming for things like payload and attributes. This means staying away from gcloud specific naming. So instead we could name it `cloud_filter_by`

rele/client.py
```diff
@@ -58,17 +58,22 @@ def create_subscription(self, subscription, topic):
...
+                request = {
+                    "name": subscription_path,
+                    "topic": topic_path,
+                    "ack_deadline_seconds": self._ack_deadline,
+                }
+
+                if subscription.filter_expression:
+                    request["filter"] = subscription.filter_expression
+
+                self._client.create_subscription(request=request)
```
tobami, line 76: is this 2.0 syntax?

EmilioCarrion: This is an alternative it always had apparently, I can't find if it's something new. You can choose both ways of invoking the method.

docs/guides/filters.rst
EmilioCarrion: I'd like something more like `backend_filter_by` as we can implement more backends in the future and it is a naming well established.

tobami: mmm, my mind was thinking more in `local_filter` vs `cloud_filter`. However if we think about a redis or rabbitmq broker, "cloud" wouldn't make sense. What about `broker_filter_by`?

xavierparedesruiz (approved):

jonasae (approved):

docs/guides/filters.rst
```diff
@@ -2,8 +2,9 @@ Filtering Messages
 ==================
 
 Filter can be used to execute a subscription with specific parameters.
-There's two types of filters, global or passing a filter_by parameter in the
-subscription.
+There's three types of filters, global, by passing a filter_by parameter in the
```
daniel-ruiz, line 5:
```suggestion
There are three types of filters, global, by passing a filter_by parameter in the
```

daniel-ruiz (approved):

antoniobusrod (commented):
What do you think to add new `backend_filter_by` support for the scenario where the topic does not exist?

I have seen that when `NotFound` exception is raised, the subscription is created without the new `filter` attribute.

EmilioCarrion:
> I have seen that when `NotFound` exception is raised, the subscription is created without the new `filter` attribute.

If the topic does not exist how could the subscription be created? 🤔

EmilioCarrion:
> > I have seen that when `NotFound` exception is raised, the subscription is created without the new `filter` attribute.
> 
> If the topic does not exist how could the subscription be created? 🤔

Oh I see what you mean! Fixing it

antoniobusrod (approved):

PR #209, opened 2021-09-10
Ignore subscriptions without topic defined

adrymyry:
### :tophat: What?

Ignore subscriptions that do not define a topic name.

### :thinking: Why?

Autodiscovery support for class-based Subscriptions (https://github.com/mercadona/rele/issues/161) opens the possibility to create generic classes with shared behavior between subclasses.

```
class GenericClass(Subscription):
    topic = None

    def __init__(self):
        self._func = self.callback_func
        super().__init__(self._func, self.topic)

    def callback_func(self, data, **kwargs):
        return data["type"]

class ClassA(GenericClass):
    topic = "topic-for-A-messages"
    
class ClassB(GenericClass):
    topic = "topic-for-B-messages"
```

`GenericClass` is a subclass of `Subscription` so it will be loaded in the autodiscovery process. However, it does not seem reasonable since it will not be subscribed to any topic.

### DIFF
--- a/rele/config.py
+++ b/rele/config.py
@@ -94,7 +94,7 @@ def load_subscriptions_from_paths(sub_module_paths, sub_prefix=None, filter_by=N
             attribute = getattr(sub_module, attr_name)
 
             subscription = subscription_from_attribute(attribute)
-            if not subscription:
+            if not subscription or subscription.topic is None:
                 continue
             if sub_prefix and not subscription.prefix:
                 subscription.set_prefix(sub_prefix)
--- a/tests/subs.py
+++ b/tests/subs.py
@@ -17,6 +17,21 @@ class ClassBasedSub(Subscription):
         return data["id"]
 
 
+class GenericClassBasedSub(Subscription):
+    topic = None
+
+    def __init__(self):
+        self._func = self.callback_func
+        super().__init__(self._func, self.topic)
+
+    def callback_func(self, data, **kwargs):
+        return data["id"]
+
+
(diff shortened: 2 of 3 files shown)


antoniobusrod:
What do you think to raise a warning instead of skipping these subscriptions? I think we should be explicit and generic about the content which should be exist in the `subs` modules.

andrewgy8 (commented):
AFAIK class based subscriptions are not well documented. It might be worth adding a simple example, in the case Relé will be supporting them as a first class feature.

rele/config.py
```diff
@@ -94,7 +94,7 @@ def load_subscriptions_from_paths(sub_module_paths, sub_prefix=None, filter_by=N
             attribute = getattr(sub_module, attr_name)
 
             subscription = subscription_from_attribute(attribute)
-            if not subscription:
+            if not subscription or subscription.topic is None:
```
tobami, line 97: I think this responsibility should fall into `subscription_from_attribute`, which already returns None if a module attribute is not of the right type. This would be a new case, which is, it is a subclass of `Subscription`, but it is "abstract".
That also allows us to log only in that case. Either a warning as @antoniobusrod suggests (thought I wonder whether that is not too aggressive), or an info or even debug (which I think it's more appropriate if abstract subscriptions are documented and valid)

tobami (commented):
As @andrewgy8 points out, class-based subscriptions are not well documented. That is not this PR's fault, but it does add even more behaviour on top of that functionality, so it makes the need to document it greater.

jonasae:
PR discontinued.

PR #212, opened 2021-12-09
[CHANGED] TimeotError from publisher

Endika:
### :tophat: What?

On [Sentry](https://sentry.prod.monline/mdona-production/delivery-api/issues/14651) register `TimeoutError` only changes the exception to catch 
```
from google.cloud.pubsub_v1.exceptions import TimeoutError
# TO
from google.cloud.pubsub_v1.publisher.exceptions import TimeoutError
```

### :thinking: Why?
If check [Sentry](https://sentry.prod.monline/mdona-production/delivery-api/issues/14651) the exception does not capture and does not execute correctly the exception code. This TimeoutError throws exception from `publisher`.
```
try:
except TimeoutError as e:
    ...
```

### :link: Related issue

[Sentry](https://sentry.prod.monline/mdona-production/delivery-api/issues/14651)

### DIFF
--- a/rele/client.py
+++ b/rele/client.py
@@ -2,12 +2,12 @@ import json
 import logging
 import os
 import time
+from concurrent.futures import TimeoutError
 from contextlib import suppress
 
 import google.auth
 from google.api_core import exceptions
 from google.cloud import pubsub_v1
-from google.cloud.pubsub_v1.exceptions import TimeoutError
 
 from rele.middleware import run_middleware_hook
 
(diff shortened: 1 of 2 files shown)


tobami (commented):
Great to see this being addressed! I imagine this broke when upgrading from v1 to v2 of the google pubsub library.

If we look at the documentation https://cloud.google.com/pubsub/docs/quickstart-client-libraries
`TimeError` is imported from the `concurrent.futures` module, shouldn't we do that instead?

tobami (approved):

PR #218, opened 2022-04-08
[Added] Added a Logging Middleware that does not truncate mesage payload.

jorchube:
### :tophat: What?

Added a `VerboseLoggingMiddleware` that mimics the output format of the current `LoggingMiddleware` but does not truncate the message payload on the post_process and pre_publish failure hooks.

### :thinking: Why?

The current `LoggingMiddleware` relies on the `__repr__()` method of the google pubsub message, which truncates the message payload. This, in turn, means that any message payload log will have that payload truncated.

The `VerboseLoggingMiddleware` implementation proposed in this PR adheres to the exact same behavior and formatting of the existing `LoggingMiddleware`, but enabling non-truncated message payload logs.


daniel-ruiz (approved):
It looks good to me 👍

Issue #222, opened 2022-06-20
Ability to setup the subscriber with dead-lettering and exponential retries


aymaneMx:
For now, there's no way to pass other args to the subscriber [request](https://github.com/mercadona/rele/blob/master/rele/client.py#L79-L83); when creating a subscriber, 

would you consider introducing this feature by either defining the `dead_letter_policy ` and the `retry_policy` in each subscriber or having a default one in the settings?

elreplicante:
Retry policies were added in v1.8.0: you can set a global `DEFAULT_RETRY_POLICY` or a per-subscription `retry_policy` in the `@sub` decorator. Dead letter policy support is being tracked in #301, so closing this one in favor of that issue.

PR #223, opened 2022-06-23
allow create a subscriber with dead-lettering & exponential retries

aymaneMx:
### :tophat: What?

allow create a subscriber with dead-lettering & exponential retries

### :thinking: Why?

we need a way to configure those 2 parameters programmatically 

### :link: Related issue

Fix https://github.com/mercadona/rele/issues/222


rele/worker.py
```diff
@@ -144,6 +152,13 @@ def create_and_run(subs, config):
         config.credentials,
         config.ack_deadline,
         config.threads_per_subscription,
+        config.retry_policy,
+        {
+            "dead_letter_topic_id": config.dead_letter_topic_id,
+            "max_delivery_attempts": config.max_delivery_attempts,
+        }
+        if config.dead_letter_topic_id
+        else None,
```
andrewgy8, line 161:
```suggestion
       config.dead_letter
```
Id suggest that the config should be handling the deadletter, and know all about how to set it up.

docs/settings.rst
```diff
@@ -160,3 +166,33 @@ library behavior, please set this value to 10.
...
 Boolean function that applies a global filter on all subscriptions.
 For more information, please see `Filtering Messages section <https://mercadonarele.readthedocs.io/en/latest/guides/filters.html#global-filter>`.
+
+
+``DEFAULT_RETRY_POLICY``
+----------------------------
+
+**Optional**
+
+Dictionary with two keys: `minimum_backoff` and `maximum_backoff`, that specifies how Pub/Sub retries message delivery for all the subscriptions.
```
jonasae, line 176:
```suggestion
Dictionary with two keys: `minimum_backoff` and `maximum_backoff`, that specifies how Pub/Sub retries message delivery for all the subscriptions in seconds.
```

docs/settings.rst
```diff
@@ -160,3 +166,33 @@ library behavior, please set this value to 10.
...
+
+    If not set, the default retry policy is applied. This generally implies that messages will be retried as soon as possible for healthy subscribers. RetryPolicy will be triggered on NACKs or acknowledgement deadline exceeded events for a given message.
```
jonasae, line 178:
```suggestion
    If not set, the default retry policy is applied, meaning a minimum backoff of 10 seconds and a maximum backoff of 60 seconds. 
    This generally implies that messages will be retried as soon as possible for healthy subscribers. RetryPolicy will be triggered on NACKs or acknowledgement deadline exceeded events for a given message.
```

jonasae:
We are splitting and making some changes from this PR.
Thanks for collaborating ;-)

PR #228, opened 2022-08-29
fix(logging): subscription_message type was different

Maks3w:
### :tophat: What?

Unify `subscription_message` data types using always an `String` 

### :thinking: Why?

After call `post_publish_failure` `subscription_message` contains a `dict` with the original data sent
After call `post_process_message_failure()` `subscription_message` contains a `Message` object

So at the end the library publish a `subscription_message` with different data types.

### DIFF
--- a/rele/contrib/logging_middleware.py
+++ b/rele/contrib/logging_middleware.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import time
 
@@ -70,7 +71,7 @@ class LoggingMiddleware(BaseMiddleware):
                     "name": "publications",
                     "data": {"agent": self._app_name, "topic": topic},
                 },
-                "subscription_message": message,
+                "subscription_message": json.dumps(message),
             },
         )
 
@@ -112,7 +113,7 @@ class LoggingMiddleware(BaseMiddleware):
                         subscription, message, "failed", start_time
                     ),
                 },
-                "subscription_message": message,
+                "subscription_message": str(message),
             },
         )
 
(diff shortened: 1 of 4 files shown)


daniel-ruiz (approved):

rele/contrib/logging_middleware.py
```diff
@@ -70,7 +71,7 @@ def post_publish_failure(self, topic, exception, message):
                     "name": "publications",
                     "data": {"agent": self._app_name, "topic": topic},
                 },
-                "subscription_message": message,
+                "subscription_message": json.dumps(message),
```
jonasae, line 74: Why we are not being consistent when casting to string?

`json.dumps(message)` vs `str(message)` in line 116

jonasae: I suggest using `message.data.decode("utf-8")`

Maks3w: The difference between json.dumps here is because "message" is in fact the raw "data" developer send when invoke publish(). In this case we don't know if that raw data can be cast to str so we have choose the json representation as we do before send the payload to the google's backend

However in the case of str this is because "message" is a Message object from the google library. In this case the Message object can be safely cast to str because it contains a __repr__ method.

Hope it helps to clarify the reasoning behind of this design decission

Maks3w: About the `decode("utf-8")` just mention this is not necessary as the value may not be str and the library does not encode the value before call this webhook.

The value is encoded but before send to the google's backend so it has to be decoded before consume the payload

Issue #229, opened 2022-09-28
Add subscription option to disable json parsing


csaroff:
Currently rele's subscription model only supports consuming messages that are json deserializable.

Unfortunately some of GCP's core services publish non-json messages. For example, the gcp dicom store publishes notifications to a topic when a new dicom instance was inserted, but the message field is just the dicom instance path. When attempting to consume these messages through rele, we get the following error.

```python
Exception raised while processing message for dicom_uploads - handle_upload: JSONDecodeError
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/rele/subscription.py", line 112, in __call__
    data = json.loads(message.data.decode("utf-8"))
  File "/usr/local/lib/python3.9/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/usr/local/lib/python3.9/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/lib/python3.9/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

Can we add an option to disable deserialization of these messages? Something like:

```python
from rele import sub

@sub(topic='my-topic', deserialize_message=False):
def handle_upload(data, **kwargs):
    print('Handling data', data)
```

andrewgy8:
Another extension to this is something where the Callback deserialize function could be configurable. 

Something like:

```
def deserialize_func(message):
	return None

@sub(topic='my-topic', deserialize=deserialize_func):
def handle_upload(data, **kwargs):
    print('Handling data', data)
```

csaroff:
Yeah, agreed. That's better

PR #230, opened 2022-09-30
Add support for arbitrary deserializers

csaroff:
### :tophat: What?

Allow custom deserializers other than the default json deserializer.

### :thinking: Why?

Some of GCP's pub/sub notifications have messages that are simple strings which can't be deserialized as json.

For example, notifications that a record was inserted into the dicom store comes through with a string indicating the path to that record and nothing more.

### :link: https://github.com/mercadona/rele/issues/229


rele/subscription.py
```diff
@@ -109,8 +122,8 @@ def __call__(self, message):
         start_time = time.time()
 
         try:
-            data = json.loads(message.data.decode("utf-8"))
-        except json.JSONDecodeError as e:
+            data = self._subscription._deserialize(message)
+        except Exception as e:
```
csaroff, line 126: I'm slightly concerned about catching arbitrary exceptions here. Feedback welcome

rele/subscription.py
```diff
@@ -145,7 +158,14 @@ def __call__(self, message):
...
 
 
-def sub(topic, prefix=None, suffix=None, filter_by=None, backend_filter_by=None):
+def sub(
+    topic,
+    prefix=None,
+    suffix=None,
+    filter_by=None,
+    backend_filter_by=None,
+    deserialize=lambda x: json.loads(x.data.decode("utf-8")),
```
andrewgy8, line 167:
```suggestion
    deserialize=lambda x: json.loads(x.data.decode("utf-8")),
```

Im thinking about the api for this. Contrary to what I recommended before, I suggest this to be `deserializer` instead. As its a thing and not a verb. WDYT?

csaroff: Agreed. deserialize sounds like a boolean

jonasae:
What do you think about changing the proposal to rely on an interface?

This would provide some benefits like:
- Anyone could provide a particular deserializer implementation
- `deserializer.deserialize()` could rise a `DeserializingException` and then avoid catching a generic exception
- Avoid passing a lambda as a default argument
- As we will be calling a method, we don't need to check if its a callable anymore

```
class DeserializingException(Exception)

class JsonDeserializer:
  def deserialize(message): 
    try:
      return json.loads(x.data.decode("utf-8"))
    except json.JSONDecodeError as e:
      raise DeserializingException(e) 
```

Then in the subscription's constructor would look like this: 
```
def __init__(
        self,
        func,
        topic,
        prefix="",
        suffix="",
        filter_by=None,
        backend_filter_by=None,
        deserializer=JsonDeserializer,
    ):
        ...
        self._deserializer = deserializer
        ...
```

And finally `Callback.__call__`:
```
def __call__(self, message):
       ...
        try:
            data = self._subscription._deserializer.deserialize(message)
        except DeseralizingException as e:
          ...
```

csaroff:
I like the exception handling where we catch a DeserializationException internally and ask users to raise that exception within their deserializer.

I don't _really_ see the point of wrapping it in a class though. Seems slightly more complicated and I don't see what it adds.

jonasae:
> I like the exception handling where we catch a DeserializationException internally and ask users to raise that exception within their deserializer.
> 
> I don't _really_ see the point of wrapping it in a class though. Seems slightly more complicated and I don't see what it adds.

Setting aside the benefits of using interfaces when injecting dependencies that might have different implementations and must fulfil a contract,  It is slightly more complicated.

At least we are removing the responsibility of checking if the passed argument is a callable (the `_init_deserializer` method won't be needed).

In terms of lines of code, both solutions are similar ;-)

andrewgy8:
Id say both solutions are identical. In the end they are callables. You could imagine passing a function, class or lambda in there would be perfectly feasible. And they would all work. 

The other argument to keeping it a generic callable rather than a specific interface is the parity with the filter_by argument in the sub decorator. Which can also be any callable.

csaroff:
> At least we are removing the responsibility of checking if the passed argument is a callable (the _init_deserializer method won't be needed).

Maybe I'm missing something, but are we?

If someone passed in a string or an int as a deserializer, wouldn't they just get an exception when we try to call their deserialize method?

```python
data = self._subscription._deserializer.deserialize(message)
```

jonasae:
> Id say both solutions are identical. In the end they are callables. You could imagine passing a function, class or lambda in there would be perfectly feasible. And they would all work.
> 
> The other argument to keeping it a generic callable rather than a specific interface is the parity with the filter_by argument in the sub decorator. Which can also be any callable. https://mercadonarele.readthedocs.io/en/latest/guides/filters.html#filter-by-parameter

Good point. For consistency sake, I'm buying it. 
But please, don't pass a lambda directly in the definition as a default callable.

jonasae:
> Maybe I'm missing something, but are we?

> If someone passed in a string or an int as a deserializer, wouldn't they just get an exception when we try to call their deserialize method?

No, you are not. In fact, it's me who missed the parenthesis in the proposal. I meant to pass an instance, not a class name.

jonasae (approved):

csaroff:
Looks like master received a force-push last month. I rebased my branch so I think the changes should be good to go!

PR #234, opened 2022-10-22
[add] PUBLISHER_BLOCKING setting

### LINKED ISSUE #202: Global blocking setting for publishing
Ability to set `blocking=True|False` globally for a project.

In many cases it is desirable to default to `True`. For example, a cron job or a worker may finish before actual publishing in the async thread is finished, hence losing messages.

Maks3w:
### :tophat: What?

Allow customization of the default value of the blocking parameter when publishing new messages. By default is `False` for Backward Compatibility preservation.

Adds documentation for this new setting.

### :thinking: Why?

As a developer, I may prefer a `blocking=True` by default so I have full quarantine that my messages have been correctly published.

### :link: Related issue

Fix #202

### DIFF
--- a/rele/config.py
+++ b/rele/config.py
@@ -4,7 +4,12 @@ import warnings
 
 from google.oauth2 import service_account
 
-from .client import DEFAULT_ACK_DEADLINE, DEFAULT_ENCODER_PATH, get_google_defaults
+from .client import (
+    DEFAULT_ACK_DEADLINE,
+    DEFAULT_BLOCKING,
+    DEFAULT_ENCODER_PATH,
+    get_google_defaults,
+)
 from .middleware import default_middleware, register_middleware
 from .publishing import init_global_publisher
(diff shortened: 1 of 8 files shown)


andrewgy8 (approved):

rele/config.py
```diff
@@ -4,7 +4,12 @@
 
 from google.oauth2 import service_account
 
-from .client import DEFAULT_ACK_DEADLINE, DEFAULT_ENCODER_PATH, get_google_defaults
+from .client import (
+    DEFAULT_ACK_DEADLINE,
+    DEFAULT_BLOCKING,
+    DEFAULT_ENCODER_PATH,
+    get_google_defaults,
```
daniel-ruiz, line 11: `isort` is complaining about the trailing comma in the last import statement. Please remove it in order for the CI pipeline to pass

```suggestion
    get_google_defaults
```

PR #235, opened 2022-10-23
[travis] PSQL is not required for tests execution

Maks3w:
### :tophat: What?

Remove PostgreSQL service from Travis pipeline

### :thinking: Why?

* It's not used
* It could give a false sensation of security by believing there are tests checking database features (Django DB)


jonasae:
As Travis has been removed as CI, this is not needed anymore.

PR #243, opened 2023-04-24
Add exponential backoff

jonasae:
### :tophat: What?

Add retry policy to the subscriptions (exponential backoff).

### :thinking: Why?

This feature allows to configure either a global (per project) retry policy or at subscription level, allowing us leveraging how  we want the subscriptions to behave (retry wise) when worker fails processing a message.

[More info about the feature here](https://cloud.google.com/pubsub/docs/handling-failures).

### :link: Related issue

#222

### DIFF
--- a/rele/retry_policy.py
+++ b/rele/retry_policy.py
@@ -12,8 +12,8 @@ class RetryPolicy:
     :param maximum_backoff: int Accepts values greater than minimum_backoff.
     """
 
-    minimum_backoff: None
-    maximum_backoff: None
+    minimum_backoff: int
+    maximum_backoff: int
 
     def __init__(self, minimum_backoff, maximum_backoff):
         self._guard_against_wrong_parameters(minimum_backoff, maximum_backoff)
--- a/rele/retry_policy.py
+++ b/rele/retry_policy.py
@@ -1,0 +1,23 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class RetryPolicy:
+    minimum_backoff: None
+    maximum_backoff: None
+
+    def __init__(self, minimum_backoff, maximum_backoff):
+        self._guard_against_wrong_parameters(minimum_backoff, maximum_backoff)
+
+        self.minimum_backoff = minimum_backoff
+        self.maximum_backoff = maximum_backoff
+
+    def _guard_against_wrong_parameters(self, minimum_backoff, maximum_backoff):
+        if minimum_backoff == 0:
+            raise ValueError("minimum_backoff must be greater than 0")
+
+        if maximum_backoff == 0:
+            raise ValueError("maximum_backoff must be greater than 0")
+
+        if minimum_backoff > maximum_backoff:
+            raise ValueError("minimum_backoff should be less than maximum_backoff.")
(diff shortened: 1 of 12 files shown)


tests/test_subscription.py
```diff
@@ -361,3 +389,15 @@ def test_logs_warning_when_function_not_in_subs_module(self, caplog):
             "Subscription function tests.test_subscription.<lambda> is outside a subs "
             "module that will not be discovered." in caplog.text
         )
+
+    def test_retry_policy_is_applied_when_specified(self):
+        subscription = sub(
+            topic="topic",
+            prefix="rele",
+            retry_policy={"minimum_backoff": 1, "maximum_backoff": 10},
```
EmilioCarrion, line 397: I would create an object RetryPolicy to avoid using dictionaries on the subscription definition. It would be a more robust API and avoid mistakes and validation errors.
So instead of:
`@sub(..., retry_policy={"minimum_backoff": 1, "maximum_backoff": 10})`
It would be:
`@sub(..., retry_policy=RetryPolicy(1, 10))`
Or something in those lines.

jonasae: I like the idea, but then I realised how complicated would be to decorate a subscription, as a RetryPolicy needs to be properly instantiated. 
Mind that the type of the arguments of the RetryPolicy constructor are [duration_pb2](https://cloud.google.com/python/docs/reference/pubsub/latest/google.cloud.pubsub_v1.types.RetryPolicy), which also require to be properly instantiated.

It would look close to this sample:
```
@sub(..., retry_policy=RetryPolicy(
        minimum_backoff=duration_pb2.Duration(seconds=10),
        maximum_backoff=duration_pb2.Duration(seconds=50),)
def decorated_subscription():
        pass
```

Furthermore, in the host project we need this setting:
```
...
    "DEFAULT_RETRY_POLICY": {"minimum_backoff": 5, "maximum_backoff": 30}
...
```

Shall we then instantiate the RetryPolicy also in the settings file?


To improve the robustness and keep the decorator and settings as light as possible, I have two alternatives besides the current implementation:
1. Get the `minimum_backoff` and `maximum_backoff` as separated arguments, which both must be present (or not).
Sample:
```
@sub(..., minimum_retry_backoff=10, maximum_retry_backoff=50)
```
2. Implement a helper `RetryPolicy` builder.
Sample:
```
@sub(..., RetryPolicy.generate_from("minimum_backoff"=5, "maximum_backoff"=30))
```

What do you think @EmilioCarrion ?

jonasae: After clarifying that the `RetryPolicy` would be a rele's object, we are embracing Emilio's suggestion, which also allows us to delegate the validation to it.

rele/retry_policy.py
```diff
@@ -0,0 +1,32 @@
...
+    """A RetryPolicy object encapsulates the validation rules.
+
+    Defines retry policy settings and ensures the values are correct.
+    If provided values are wrong, a ValidationError is raised.
+
+    :param minimum_backoff: int Accepts values greater than 0
+    :param maximum_backoff: int Accepts values greater than minimum_backoff.
+    """
+
+    minimum_backoff: None
```
andrewgy8, line 15:
```suggestion
    minimum_backoff: Optional[Int]
```
Should this be an int?

jonasae: Yep, but not optional. Thanks!

EmilioCarrion (approved):

PR #253, opened 2023-05-08
Add logs when a worker is stopped

jsansaloni:
### :tophat: What?
Add logs to trace how the worker is being stopped

### :thinking: Why?
To have more data to understand what is happening

### DIFF
--- a/rele/worker.py
+++ b/rele/worker.py
@@ -88,11 +88,19 @@ class Worker:
         :param signal: Needed for `signal.signal <https://docs.python.org/3/library/signal.html#signal.signal>`_  # noqa
         :param frame: Needed for `signal.signal <https://docs.python.org/3/library/signal.html#signal.signal>`_  # noqa
         """
+
+        logger.info(f"[rele] {signal} received, stopping workers")
+
         run_middleware_hook("pre_worker_stop", self._subscriptions)
         for future in self._futures.values():
             future.cancel()
 
+        logger.info("[rele] future cancelled")
+
         run_middleware_hook("post_worker_stop")
+
+        logger.info("[rele] worker stopped")
+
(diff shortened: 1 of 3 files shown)


rele/worker.py
```diff
@@ -88,11 +88,19 @@ def stop(self, signal=None, frame=None):
...
+
         run_middleware_hook("pre_worker_stop", self._subscriptions)
         for future in self._futures.values():
             future.cancel()
 
+        logger.info("[rele] future cancelled")
+
         run_middleware_hook("post_worker_stop")
+
+        logger.info("[rele] worker stopped")
```
jonasae: Are we sure that self._futures is always populated? It's going to log either it has cancelled a future or not.
Also I think that we are poluting the method with intermediate logs.

I have checked the `run_middleware_hook` function and it doesn't have any log.

What do you think about logging inside `run_middleware_hook` at the beginning and the end, including the `hook_name`?

rele/worker.py
```diff
@@ -88,11 +88,19 @@ def stop(self, signal=None, frame=None):
...
         run_middleware_hook("pre_worker_stop", self._subscriptions)
         for future in self._futures.values():
             future.cancel()
 
+        logger.info("Rele cancel all futures")
+
         run_middleware_hook("post_worker_stop")
+
+        logger.info("Rele worker stopped")
+
```
jonasae, line 103: Are we sure that self._futures is always populated? It's going to log either it has cancelled a future or not.
Also I think that we are poluting the method with intermediate logs.

I have checked the run_middleware_hook function and it doesn't have any log.

What do you think about logging inside run_middleware_hook at the beginning and the end, including the hook_name?

PR #255, opened 2023-07-03
[Changed] Detect subs module at any folder level

albertorausell:
### :tophat: What?
We have changed the `discover_subs_modules` function in order to find every subs module independently of the file system level where it is

### :thinking: Why?
Because we want to integrate the hexagonal architecture and subs should be in infrastructure module


EmilioCarrion (approved):

EmilioCarrion:
We should make a pre-release candidate as beta and deeply test it to avoid missing subs in production environments

PR #257, opened 2023-07-05
[Added] Check if subs have same memory address

alcaptar:
### :tophat: What?

Checking if imported subs have the same memory address.
Also, not duplicating the paths of the subs when discovering them.

### :thinking: Why?

To avoid raising a duplicate subscription exception when actually is the same subscription, just being imported several times.

### :link: Related issue

[Related PR](https://github.com/mercadona/rele/pull/255)


rele/config.py
```diff
@@ -83,6 +83,20 @@ def setup(setting=None, **kwargs):
...
 
 
+def subscriptions_have_same_memory_address(a_subscription, another_subscription):
+    return id(a_subscription) == id(another_subscription)
+
+
+def subscription_already_registered(subscriptions, current_subscription):
+    registered_subscription = subscriptions.get(current_subscription.name)
+    if not registered_subscription:
+        return False
```
EmilioCarrion, line 93: Is this part really needed?

rele/config.py
```diff
@@ -83,6 +83,20 @@ def setup(setting=None, **kwargs):
     return config
 
 
+def subscriptions_have_same_memory_address(a_subscription, another_subscription):
```
EmilioCarrion, line 86:
```suggestion
def is_same_susbscription(a_subscription, another_subscription):
```
I would include the `is` verb in the method naming

rele/config.py
```diff
@@ -83,6 +83,20 @@ def setup(setting=None, **kwargs):
...
+    return id(a_subscription) == id(another_subscription)
+
+
+def subscription_already_registered(subscriptions, current_subscription):
```
EmilioCarrion, line 90:
```suggestion
def is_subscription_already_registered(subscriptions, current_subscription):
```
I would include the `is` verb in the method naming

EmilioCarrion (approved):

Issue #266, opened 2024-01-18
Possibility to specify packages for subscriptions discovery


awaistkd:
Right now rele discovers the subscriptions from the packages in the [current path](https://github.com/mercadona/rele/blob/9dc6a68ae06cc710dda5bc3025d917b50019bd7b/rele/discover.py#L44). 

We want the ability to discover the subscriptions from a package that is installed by the pip. For example:

```python
RELE = {
     ...
    "DISCOVER_SUBSCRIPTIONS_IN": [".",  "package_installed_using_pip"]
}
```

elreplicante:
This is now supported: with the standalone worker, `rele-cli run --third-party-subscriptions my_package.subs another_package.subs` (added in #269); with Django, any pip-installed app added to `INSTALLED_APPS` that ships a `subs` module is auto-discovered. Closing as resolved.

PR #269, opened 2024-01-19
Added --third-party-subscriptions argument in rele-cli run command

awaistkd:
### :tophat: What?
Added `--third-party-subscriptions` argument in `rele-cli run` command which will allow us to pass subscriptions from the third-party packages.

### :thinking: Why?

When we run `python manage.py runrele` command, the Rele is able to automatically discover the subscriptions from the installed third-party packages if they are added to the `INSTALLED_APPS`. 

We need the possibility to run the subscriptions from a third-party package with the `rele-cli run` command .

### :link: Related issue

Related issue --> #266


rele/discover.py
```diff
@@ -29,9 +31,9 @@ def _import_settings_from_path(path):
         return importlib.import_module(path)
 
 
-def sub_modules(settings_path=None):
+def sub_modules(settings_path=None, third_party_packages=None):
     """
-    In the current PYTHONPATH, we can traverse all modules and determine if they
+    In the current python path, we can traverse all modules and determine if they
```
igarridot, line 36:
```suggestion
    In the current directory, we can traverse all modules and determine if they
```

tests/commands/test_rele_cli.py
```diff
@@ -0,0 +1,10 @@
+from rele.__main__ import run_worker
+
+
+class TestReleCli:
+    def test_rele_cli_run(self, mock_worker):
+        run_worker("tests.settings", ["sample_pypi_package.subs"])
```
jBonoraW, line 6: What about add test for the validation of the module.
Could be easy as adding "sample_pypi_package.no_subs" in this happy case or add anew test like:
```
    def test_ignores_non_valid_third_party_subs(self, mock_worker):
        run_worker("tests.settings", ["sample_pypi_package.no_subs"])

        mock_worker.assert_called()
```

jBonoraW (approved):

PR #287, opened 2025-01-24
Allow create topics on multiple Google Pub/Sub regions

mvillarjuan:
### :tophat: What?

Allow creating topics on multiple Google Pub/Sub regions

### :thinking: Why?

We want topics to be created on multiple Google Pub/Sub regions to improve reliability and GDPR compliance

### :link: Related issue
#288

### DIFF
--- a/rele/client.py
+++ b/rele/client.py
@@ -55,7 +55,9 @@ class Subscriber:
         self._gc_project_id = gc_project_id
         self._ack_deadline = default_ack_deadline or DEFAULT_ACK_DEADLINE
         self.credentials = credentials if not USE_EMULATOR else None
-        self._message_storage_policy = self._normalize_storage_policy(message_storage_policy)
+        self._message_storage_policy = self._normalize_storage_policy(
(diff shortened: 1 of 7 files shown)


Sermanes (approved):

vicmarbev (approved):

rele/client.py
```diff
@@ -95,11 +97,33 @@ def _create_topic(self, topic_path):
...
+        """
+        Normalizes the storage policy to accept either a string or a list.
+        If it is a string, it converts it into a list with a single element.
+        """
+
+        if isinstance(policy, str):
+            # Warning log for future compatibility
+            print(
+                "Warning: `message_storage_policy` as a string is deprecated. Use a list of regions instead."
+            )
```
EmilioCarrion, line 115:
```suggestion
            warnings.warn(
                "`message_storage_policy` as a string is deprecated. Use a list of regions instead.",
                DeprecationWarning,
            )
```
https://docs.python.org/3.12/library/warnings.html

PR #302, opened 2026-05-22
[Fixed] Avoid returning from finally in worker connection check

### LINKED ISSUE #299: Python 3.14 SyntaxWarnings: 'return' in a 'finally' block
## Description
When running with Python 3.14, Rele emits multiple SyntaxWarning messages about return statements inside finally blocks:

```
/usr/local/lib/python3.14/site-packages/rele/worker.py:38: SyntaxWarning: 'return' in a 'finally' block
  return result
```

## Why This Matters
Using return inside a finally block is problematic because:

1. It silently swallows any exceptions that were raised in the try block
2. It masks the original exception, making debugging difficult
3. Python 3.14 now explicitly warns about this pattern

## Affected Files
`rele/rele/worker.py (line 38)`

## Proposed Fix
Refactor the code to move the `return` statements outside of the `finally` blocks while preserving the intended behavior.

## Environment
- Python version: 3.14
- rele version: latest (main branch)

Mirochill:
### :tophat: What?

- Move `check_internet_connection()`'s `return result` outside the `finally` block so Python 3.14 no longer warns about `return` inside `finally`.
- Keep the socket close in `finally`.
- Add regression coverage for the success and socket-error paths.
- Add myself to `AUTHORS.md` as requested by `CONTRIBUTING.md`.

### :thinking: Why?

Returning from `finally` can mask exceptions raised in the `try` block, and Python 3.14 now emits a `SyntaxWarning` for this pattern. Returning after `finally` keeps the current boolean result while avoiding the warning.

Validation:
- `git diff --check`
- `git diff --cached --check`
- `git diff --check HEAD~1..HEAD`
- Not run locally

### :link: Related issue

Fix #299

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -16,3 +16,4 @@ Thank you to all the contributors to Relé!
 * Chirag Jain (@chirgjin-les)
 * Jordi Chulia (@jorchube)
 * Emilio Carrión (@EmilioCarrion)
+* Miro (@Mirochill)
--- a/rele/worker.py
+++ b/rele/worker.py
@@ -35,7 +35,7 @@ def check_internet_connection(remote_server):
         logger.exception("Check internet connection error", error)
     finally:
         sock.close()
-        return result
+    return result
 
 
 class Worker:
--- a/tests/test_worker.py
+++ b/tests/test_worker.py
@@ -1,3 +1,4 @@
+import socket
 import time
 from concurrent import futures
 from concurrent.futures._base import FINISHED
@@ -13,7 +14,7 @@ from rele import Subscriber, Worker, sub
 from rele.middleware import register_middleware
 from rele.retry_policy import RetryPolicy
 from rele.subscription import Callback
-from rele.worker import NotConnectionError, create_and_run
+from rele.worker import NotConnectionError, check_internet_connection, create_and_run
 
 
 @sub(topic="some-cool-topic", prefix="rele")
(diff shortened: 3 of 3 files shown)


elreplicante:
Thanks for this, @Mirochill — and sorry for the long silence.

The fix is now released in **[1.17.0](https://github.com/mercadona/rele/releases/tag/v1.17.0)** (on PyPI since today), so `check_internet_connection` no longer returns from inside `finally` and the Python 3.14 `SyntaxWarning` from #299 is gone.

Credit where it's due, though: **you got there first.** You opened this on 2026-05-22, and the change that actually landed (#303, merged 2026-07-22) is the identical one-liner — same `return result` moved out of the `finally`, same reasoning about exceptions being swallowed. Your PR should have been the one merged, and it wasn't because we let it sit. That's on us, not on you.

Your diff also contained something #303 did **not**, and that we still don't have:

- the regression tests (`TestCheckInternetConnection` — socket closed on the success path and on `socket.error`). The current tests around `check_internet_connection` only cover the `api_endpoint` setting, so those two paths are genuinely untested on master.
- your entry in `AUTHORS.md`.

We'll port both ourselves rather than ask you to rebase a branch this old — with you credited as the author of the tests and added to `AUTHORS.md`. No further work needed from your side.

Thanks again for picking up #299 and for including tests with the fix. It's appreciated, even if our process didn't show it this time.

elreplicante:
Done, @Mirochill — your tests are on `master` as of #330, ported verbatim, with you as `Co-authored-by` on the commit and added to `AUTHORS.md`.

One thing worth telling you: **your tests found a second bug.**

The error-path test wouldn't pass. Not because of the test — because of this, a few lines below the `finally` you were fixing:

```python
logger.exception("Check internet connection error", error)
```

`error` goes in as a `%`-format argument to a message with no placeholder. At format time that raises `TypeError`, `logging` swallows it through `handleError`, and the record is **never emitted** — so the connection-check error log has been silently disappearing since #283, exactly when connectivity fails and you'd want it. Nothing crashes, which is why nobody noticed. It only showed up because pytest's `LogCaptureHandler` re-raises formatting errors instead of swallowing them, so your test hit it head-on.

Fixed in the same PR. You wrote two tests for a one-line `finally` fix and they caught an unrelated latent bug in the same function — that's the argument for including tests with a fix, made better than we could have made it.

Thanks again. Closing this one, since everything in it has landed: the `finally` fix in 1.17.0, the tests and your credit in #330.

